PR #2, opened 2023-02-20
add content to readme

gregsheremeta:
add content to readme. Most of this content will be duplicated in ADR-0001, the ADR that establishes the process.


gregsheremeta:

PR #3, opened 2023-02-20
switch license to apache

gregsheremeta:
switch license to apache

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,674 +1,201 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
-
- Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
-
-                            Preamble
-
-  The GNU General Public License is a free, copyleft license for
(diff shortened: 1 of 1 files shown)


gregsheremeta:

accorvin:
@gregsheremeta any reason why you chose Apache specifically? We use GPLv3 elsewhere in the ODH.

gregsheremeta:
The team talked about licenses the other day, and it wasn't clear to us which one is supposed to be used globally. Ironically, it's something we'll need an ADR for.

I'm surprised to hear GPLv3 is what we commonly use, for two reasons:

1 The subprojects in opendatahub-io seem mixed. I copied the LICENSE file from https://github.com/opendatahub-io/data-science-pipelines/blob/master/LICENSE -- although that is a fork, so maybe that one was a bad one to use as an example.

2 Projects that Red Hat productizes are usually Apache 2.0 -- oVirt, OpenStack, Kubernetes, Ansible ... all Apache 2.0.

All that said, this repo is text only, so Creative Commons is perhaps the most appropriate.

cc @LaVLaS

gregsheremeta:
slack thread for further discussion: https://odh-io.slack.com/archives/C033X6XM448/p1676975429593999

PR #5, opened 2023-02-20
add ADR template

gregsheremeta:
add ADR template - just a skeleton for now


gregsheremeta:

PR #13, opened 2023-06-23
ADR for mlserving UWM

VedantMahabaleshwarkar:


VedantMahabaleshwarkar:
@etirelli @danielezonca @jgarciao @andrewballantyne Can I get a review on this ADR?

ODH-ADR-0006-use-user-workload-monitoring-mlserving.md
```diff
@@ -0,0 +1,51 @@
...
+* A Solution for long term metric retention is out of the scope for this ADR but will be covered in the technical design document for the solution. 
+ 
+## How
+
+The technical details surrounding this change require multiple considerations in terms of implementation, and it is not feasible to cover all the necessary details in an ADR document. The technical design document for this change captures the "How" behind this change. 
+
+## Open Questions
+
+1.  Enabling User Workload Monitoring requires `cluster-admin` permissions. This will the expectation that model serving metrics show up OOTB in RHODS when a model is deployed. 
+We need documentation/UX alerts to let `rhods-admins` know that they need to enable UWM in collaboration with `cluster-admins` to enable the full functionality of model serving metrics once this change is implemented. What is the best way of ensuring a smooth UX during this transition?
```
andrewballantyne, line 39: I don't believe this is 100% true. You need to be able to list Namespaces to fetch wider metrics... but for your namespace, you can do this with just a project permission (not sure how much -- just access to it? have not tested)

This is assuming I understand this correctly... `thanos-querier` queries UWM:
* `https://thanos-querier.openshift-monitoring.svc.cluster.local:9092` is accessible by the basic user
* `https://thanos-querier.openshift-monitoring.svc.cluster.local:9091` is accessible by users who can list Namespaces

[Slack conversation](https://redhat-internal.slack.com/archives/C6A3NV5J9/p1676490553300299) I had with the OpenShift folks.

Are you saying that our Model Serving in a namespace is going to require listing Namespace permissions? Because I'm not sure that's ideal.

andrewballantyne (changes requested):

ODH-ADR-0006-use-user-workload-monitoring-mlserving.md
VedantMahabaleshwarkar: @andrewballantyne all of this is correct `AFTER` the global flag for UWM is enabled. Enabling the global flag itself needs cluster-admin. 
https://docs.openshift.com/container-platform/4.13/monitoring/enabling-monitoring-for-user-defined-projects.html

andrewballantyne: Well, as it stands the Dashboard expects it to work for PVC fill-amount values. So there is that.

VedantMahabaleshwarkar: hmm interesting. So is the implication that all working RHODS installs currently would already have UWM enabled on the cluster? Also if yes, since UWM is technically under the cluster umbrella and not specific to rhods itself, should it not involve some verification/error throwing around the case where it isn't enabled? What happens in the dashboard currently if UWM is disabled by a cluster-admin?

andrewballantyne: We throw 500 errors attempting to get PVC fill values in DS Projects on the Dashboard and this alerts SRE as I understand it. I imagine our managed clusters always have UWM and I have not heard of problems in Self Managed or in ODH community.

strangiato:
So a few thoughts...

ODH/RHODS should not be responsible for enabling end user workload monitoring on the cluster.  The operator should not make changes to cluster level configurations, and it is the responsibility of the cluster admin to enable this feature.  Enabling the end user workload monitoring feature without the cluster admin being made aware could lead to degraded performance in the cluster prometheus instance or worse.

The requirement for end user workload monitoring should be well documented, and we should leverage official documentation on how to enable the feature, similar to how we document the requirements around the GPU/NFD operators.

The dashboard should make it clear when this feature is not enabled and display a message recommending the cluster admin enable it to take advantage of the monitoring capabilities or setup a separate prometheus instance, which leads to....

ODH should still enable the use of any prometheus instance.  End user workload monitoring is a great feature, but it is openshift specific.  In order to keep the project at least possible to run on other k8s clusters or a number of architectural designs that may require us to not use end user workload monitoring we should still enable using another prometheus instance.

From a dashboard perspective I would expect this to mean that an admin should be able to configure a connection to a prometheus instance if they do not want to utilize end user workload monitoring.  In this use case I would expect the end user to be 100% responsible for installing and managing the prometheus instance and they would be required to setup the connection data in ODH to allow the dashboard to read the data.  In the case we are using end user workload monitoring this connection could be setup for users automatically.

One final thought is that as we are working on setting up Pod/Service Monitoring objects, these should not be created by the dashboard when we create a ModelMesh object.  These objects should be managed by the ModelMesh controller instead and a child of the ModelMesh instance.

andrewballantyne:
This was a great informative comment @strangiato!

Today, our ODH/RHODS stack expects it to be on an OpenShift instance -- if that was not the desired case, I think there are several architectural gaps in the way Dashboard reaches out to look for k8s resources that are CRDs from OpenShift.  

Worth noting, the Dashboard creates nothing in the monitoring stack and I do not expect that to change. But you made a good point about the Dashboard should have configurations for setting up an endpoint for where to get metrics. Part of this effort, we should look to move to a system of configuration for Model Serving metrics. Having a backing OpenShift Route (name & namespace) -- or if there is a better mechanism -- to where to send requests for monitoring aspects.

If UWM is optional, we need to also consider what we are doing with the PVC "filled state" value calls we do on DS Projects.

VedantMahabaleshwarkar:
@strangiato fully agree with your points. I had originally thought of the same concerns, hence the `open questions` section of the ADR. We need to finalize a plan to document the requirement for UWM in MLServing Metrics, have some sort of awareness for all types of users (possibly in the dashboard? similar to the warning we show for DSPO if the pipelines operator isn't installed). I also agree that it shouldn't be the rhods-admin solely looking at UWM but it should be in collaboration with the cluster admin.  


> ODH should still enable the use of any prometheus instance.  

This is increasingly becoming a difficult use-case to support as managed openshift continues to move towards an approach of "No custom prometheuses please". Continuing to support this will require us to maintain multiple implementations for ODH and the downstream product, which needs further discussion if that's the route we want to take. 

@andrewballantyne I think having a configurable endpoint setup for metrics would be the best case scenario, but I'm not sure how practically achievable that is in terms of efforts. For now, what do you think about having some sort of detection mechanism to see if UWM is enabled? If we detect that it isn't enabled we could show a simple error on the metrics page that can inform the user to talk to rhods-admin + cluster-admin to enable UWM.

ODH-ADR-0006-use-user-workload-monitoring-mlserving.md
```diff
@@ -0,0 +1,52 @@
...
+| Superseded by: | N/A |
+| Tickets        | TBD |
+| Other docs:    | [Technical Design Document](https://docs.google.com/document/d/16R1dDOwZri7X_7cALuNK9lkLDF5cTJMKwHpSJRINAzA/edit) |
+
+## What
+
+This is a document intended to capture the decision to use [User Workload Monitoring](https://docs.openshift.com/container-platform/4.13/monitoring/enabling-monitoring-for-user-defined-projects.html) as the monitoring stack to generate metrics for mlserving
+## Why
+
+The decision to user User Workload Monitoring for Ml Serving metrics impacts multiple areas of work within RHODS. This ADR document acts as the high level proposal for the change as well as records the sign-off from all stakeholders. To summarize, the current monitoring stack for model serving fails to provide maintainable solutions to our monitoring requirements. Further details on the exact failures can be found in the technical design document for the solution.
```
etirelli, line 21: typo: "use"

ODH-ADR-0006-use-user-workload-monitoring-mlserving.md
```diff
@@ -0,0 +1,52 @@
...
+
+* A Solution for long term metric retention is out of the scope for this ADR but will be covered in the technical design document for the solution. 
+ 
+## How
+
+The technical details surrounding this change require multiple considerations in terms of implementation, and it is not feasible to cover all the necessary details in an ADR document. The technical design document for this change captures the "How" behind this change. 
+
+## Open Questions
+
+1.  Enabling User Workload Monitoring requires `cluster-admin` permissions. This will the expectation that model serving metrics show up OOTB in RHODS when a model is deployed.
```
etirelli, line 38: typo: "This will the expectation"

jgarciao:
> Will this support disconnected configurations?

@VedantMahabaleshwarkar I think this should be a requirement for the proposed solution. I've proposed it as a requirement in the design document

I've also proposed a requirement about compatibility with [RHODS component dependencies](https://docs.google.com/document/d/1fn0rFycl4SKukCX7_7N7NK49x4PXS4tjzmzboZKK3yY/edit?usp=sharing) (OpenShift Service Mesh, OpenShift Pipelines, …)", because the UWM documentation says 

- _In OpenShift Container Platform 4.13 you must remove any custom Prometheus instances before enabling monitoring for user-defined projects_ 

and I don't know if RHODS dependencies deploy custom prometheus instances

PR #16, opened 2023-08-28
Add ADR-Operator-0001 for odh-manifests trainsition

zdtsw:
## Description

Add ADR-Operator-0001 for odh-manifests trainsition

ref: https://github.com/opendatahub-io/opendatahub-operator/issues/338

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #17, opened 2023-09-05
Adding ADR Operator 0002 - ODH Operator Scope

etirelli:
Adding ADR 0009 detailing the reasoning and impact of the ODH Operator v2 scope.


ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+- ServiceMesh
+- Serverless
+- NFD 
+
+In order to properly manage these resources, the operator requires access and permissions across namespaces. In particular, the operator creates and/or manages four namespaces:
+
+- redhat-ods-operators
+- redhat-ods-applications
+- redhat-ods-monitoring
+- rhods-notebooks
```
VaishnaviHire, line 63: `rhods-notebooks`  is not created on installation, only created when `workbenches` component is deployed

etirelli: yes, but I think the point of the ADR doc is that the operator still creates/manages the namespace in that circunstance (even if not automatically during the installation). https://github.com/opendatahub-io/opendatahub-operator/blob/main/components/workbenches/workbenches.go#L53

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+  
+The operator also leverages the platform’s [Owner References](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) capability that ensures all dependent resources are owned by the operator and their life cycles are tracked and managed cleanly. Kubernetes however restricts the use of owner references in a way that only cluster scoped resources can own other cluster scoped resources, leading to the requirement of the operator being cluster scoped. 
+
+## Goals
+
+* The operator needs to support management of resources across namespaces
+* The operator needs to manage third party operators
```
strangiato, line 70: The term "third party operators" is a bit confusing.  "third party" could mean, "not Red Hat" (e.g. NVIDIA GPU Operator), "Not ODH" (e.g. openshift-pipelines), or "not the ODH operator", (e.g. DSP Operator).

I would recommend trying to standardize some of the language with something like "external operators" and "component operators/controllers".

Additionally, it isn't clear what "manage" means in this case.  I am guessing that you are talking about "managing" the "component operators/controllers" so that cleanup of language may help to clarify things here.

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+- Secrets
+- Service
+- NetworkPolicy
+- Route
+  
+The operator deploys the following Custom Resource Definitions:
+
+- DSCInitialization
+- DataScienceCluster
+- OdhApplication
```
strangiato, line 42: I think that the decision on which CRDs are managed/deployed by the operator and which are deployed by the components should be revisited.  This may be a topic outside of the scope of this ADR but the inclusion of the Dashboard objects on the main operator, and not other objects such as the Notebook object seems haphazard and inconsistent.

This topic of "what component CRDs get included in the operator" seems like a prime candidate to formalize as an ADR.

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+- OdhDocument
+- OdhDashboardConfig
+- OdhQuickstarts
+ 
+The operator (optionally) requires and/or manages the following dependent operators:
```
strangiato, line 47: OLM supports a dependency feature which allows OLM to install a dependent operator if it is not installed.  It may be good to clarify that these dependencies are ones that users are expected to install in order to support those specific components that require them, and is not referring the dependency features of OLM.

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+- rhods-notebooks
+  
+The operator also leverages the platform’s [Owner References](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) capability that ensures all dependent resources are owned by the operator and their life cycles are tracked and managed cleanly. Kubernetes however restricts the use of owner references in a way that only cluster scoped resources can own other cluster scoped resources, leading to the requirement of the operator being cluster scoped. 
+
+## Goals
+
+* The operator needs to support management of resources across namespaces
+* The operator needs to manage third party operators
+* The operator must track all resources it creates and manages
+* The operator must clean up resources upon uninstall
```
strangiato, line 72: "clean up all resources upon uninstall" is vague and doesn't describe the behavior I would expect of the operator.

If I have deployed a DSC, I should be able to uninstall the operator, and the DSC and the dependent resources it has deployed should remain, in an unmanaged state.

If I remove a DSC, while the operator is installed, it should remove the dependent resources created by that DSC.

If I remove a DSC, after the operator is uninstalled, the dependent resources should remain in an unmanaged state.

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+The operator leverages platform capabilities for that task, like owner references. Namespaced operators cannot own resources outside their own namespace (as per Kubernetes design). Scoping the operator to a namespace would require us to stop using owner references and implement from scratch the management of resources, using a different mechanism like labels. That would be complex, brittle, and duplicate functionality available in the platform itself.
+
+### 3. Contradictory requirements with other users and best practices
+The use of namespaced operator would prevent us from meeting the requirements of other customers and users (e.g. automated install and upgrades of resources across namespaces). 
+
+## Tradeoffs
+
+The main drawback of using a cluster scoped operator is the impossibility of running multiple operators in the same cluster (e.g., operators of different versions in different namespaces). However, this by itself is an anti-pattern. The recommendation for users that need different operator versions for different environments is to use Hypershift (or Hosted Controlpanel) to deploy one cluster for each environment.
```
strangiato, line 99: The trade off in capabilities is that you can't run multiple instances/versions of  RHODS environment in the same cluster.  Installing multiple instances of a namespace scoped operator is simply a way to accomplish that goal but it is not the objective itself.

For the recommendation, I would recommend simply managing multiple clusters, not Hypershift.  Hypershift is a pretty niche use case that adds a lot of complexity over just managing multiple OCP clusters.  I'm not sure that I would recommend it if the goal is simply "I need multiple RHODS environments".

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+- Namespaces
+- Role
+- ClusterRole
+- RoleBindings
+- ClusterRoleBindings
+- ConfigMaps
+- Secrets
+- Service
+- NetworkPolicy
+- Route
```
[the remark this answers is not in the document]
zdtsw: do we need to list all resources here? or just some example is enough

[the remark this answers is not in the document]
etirelli: Since this changes over time, I think a snapshot of the current situation is enough as a justification. I added deployments.

ODH-ADR-0009-odh-operator-scope.md
etirelli: I think we can discuss this as part of another ADR. The point of this ADR is to clarify and justify the decision to use a cluster scope for the operator.

ODH-ADR-0009-odh-operator-scope.md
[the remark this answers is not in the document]
etirelli: The goal for this ADR is to clarify and justify the need for the cluster scope for the operator. We should add enough relevant evidence driving the decision, but I don't think it needs to be comprehensive for the purpose of justifying scope. Added NVIDIA GPU and openshift-pipelines.

etirelli: OLM dependencies are mandatory, AFAIK. The requirements is for ODH/RHODS operator to only install dependencies if the component is enabled.

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+- OdhApplication
+- OdhDocument
+- OdhDashboardConfig
+- OdhQuickstarts
+ 
+The operator (optionally) requires and/or manages the following dependent operators:
+
+- KServe
+- ModelMesh
+- OdhModelController
```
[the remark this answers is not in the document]
etirelli: removed

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,116 @@
...
+- Ray
+- Codeflare
+- ServiceMesh
+- Serverless
+- NFD 
+
+In order to properly manage these resources, the operator requires access and permissions across namespaces. In particular, the operator creates and/or manages four namespaces:
```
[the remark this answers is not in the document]
etirelli: added a comment stating that the names are configurable.

ODH-ADR-0009-odh-operator-scope.md
etirelli: good points. We also had a long discussion on this particular topic today and there isn't consensus that this goal is achievable, so I am removing it for now.

ODH-ADR-0009-odh-operator-scope.md
etirelli: You are correct. However the focus on this document is to justify the need for the cluster scope of the operator. In that context, I am not sure trying to define what the clean up rules are for each type of resource would be relevant. Do you have a suggestion on how to word it better?

ODH-ADR-0009-odh-operator-scope.md
etirelli: reworded.

ODH-ADR-0009-odh-operator-scope.md
strangiato: What about this:
"The operator must remove resources created by a managed custom resource when that custom resource is removed."

ODH-ADR-0009-odh-operator-scope.md
strangiato: My point was primarily that this reads like you are talking about the dependency feature in OLM and it is important to distinguish that this is not the case.

I would be interested to discuss that as a requirement more since that is something we have actively moved away from in the past due to it causing a lot of issues.

ODH-ADR-0009-odh-operator-scope.md
VaishnaviHire: Agreed, need to be specific for ODH

ODH-ADR-0009-odh-operator-scope.md
etirelli: done

ODH-ADR-0009-odh-operator-scope.md
etirelli: done

ODH-ADR-0009-odh-operator-scope.md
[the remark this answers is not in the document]
strangiato: These namespaces are still not correct.  The ODH operator does not manage or create any of these namespaces.  

The ODH operator creates the `opendatahub` namespace and the RHODS operator creates the following.  I don't take an issue if you want to include these in the ADR but you need to explicitly state that this is for RHODS, not ODH, and you still need to describe what ODH does.

What about something like this?

```
In order to properly manage these resources, the operator requires access and permissions across 
 multiple namespaces namespaces. In particular the operator creates and manages the following namespace:

- opendatahub

Additionally, the operator provides the capability to configure which namespaces are created and managed.  The downstream RHODS operator uses this capability to configure and manage the following namespaces:

- redhat-ods-operators
- redhat-ods-applications
- redhat-ods-monitoring
- rhods-notebooks
```

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,115 @@
...
+- ServiceMesh
+- Serverless
+- NFD 
+- NVIDIA GPU
+- openshift-pipelines
+
+In order to properly manage these resources, the operator requires access and permissions across namespaces. In particular, the operator creates and/or manages the following namespaces (these are the default ODH names, but they can be changed via configuration):
+
+- opendatahub
+- openshift-operators
```
strangiato, line 63: This NS is not created or managed by the ODH operator.  This should probably just list opendatahub ns

ODH-ADR-0009-odh-operator-scope.md
```diff
@@ -0,0 +1,115 @@
...
+- Service
+- NetworkPolicy
+- Route
+  
+The operator deploys the following Custom Resource Definitions:
+
+- DSCInitialization
+- DataScienceCluster
+- OdhApplication
+- OdhDocument
```
VaishnaviHire, line 44: @etirelli We no longer deploy Dashboard CRDs through the operator

VaishnaviHire:
Should this also be moved under `operator` similar to [ADR-Operator-0001](https://github.com/opendatahub-io/architecture-decision-records/pull/16/files#diff-00b391b44327084ab70b7a4ea7eb92d7f60f2d9b03b6be6105a9d875b36a21af)

ODH-ADR-0009-odh-operator-scope.md
etirelli: done

ODH-ADR-0009-odh-operator-scope.md
etirelli: done

VaishnaviHire:
> Moved to the `operator` folder and renumbered to ADR 0004

Shouldn't this be ADR 0002?

zdtsw:
> > Moved to the `operator` folder and renumbered to ADR 0004
> 
> Shouldn't this be ADR 0002?

the file actually is called `operator/ODH-ADR-Operator-0004-operator-scope.md` :D

etirelli:
@VaishnaviHire The last PR I have seen was for the operator ADR 0003, so I assumed 0001-0003 were used. If 0002 is not used, I can rename the file.

strangiato:
lgtm

etirelli:
@VaishnaviHire I renumbered the ADR to 0002 and updated the status to approved. Let me know please if there is anything else.

PR #23, opened 2023-12-01
ODH-ADR-Operator-0004-replica-image-support

AjayJagan:
Update DSC CRD with new customization values.

## Description
Upgrade the DSC CRD to support customizing the replica count, images and modifying the resource limits/requests of individual components.

### DIFF
--- a/operator/ODH-ADR-Operator-0004-replica-image-support.md
+++ b/operator/ODH-ADR-Operator-0004-replica-image-support.md
@@ -17,36 +17,64 @@ This document explains about upgrading the DSC CRD to support customizing the re
 
 ## Why
 
-Current operator design doesnot support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields.
-NOTE: The image update is targeted for dev purposes and will not be available in production. 
+Current operator design does not support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields.
+NOTE: The image update is targeted for dev purposes and will not be available in a supported configuration. 
 
 ## Goals
 
-* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count. If a component is deploying multiple deplyoments, then something like this is preferred:
+* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count. An example yaml is given below:
   ```yaml
     spec:
   components:
      workbenches:
-          notebookController:
-               replicaCount: <>
-          kfNotebookController:
-                replicaCount: <>
+        managementState: Managed
+        notebookController:
+            replicas: 1
+        kfNotebookController:
+            replicas: 1
   ```
 * For each component(eg. codeflare, ray, kserve etc) provide an image field in the *devFlags* through which we can pass-in a custom image to be used by the controller.
-* Certain components uses other images(for eg. datascience-pipelines-controller uses `IMAGE_API_SERVER`, `IMAGES_ARTIFACT`). We need to provide support to customize these images as well in the *devFlags*
-* Control the resource limits and requests for individual components for more fine grained control.
+* Control the resource limits and requests for individual components for more fine grained control. An example yaml is given below:
+  ```yaml
+    spec:
+  components:
+    datasciencepipelines:
+      managementState: Managed
+      data-science-pipelines-operator-controller-manager:
+        resources:
+          requests:
+            cpu: 100m
+            memory: 128Mi
+          limits:
+            cpu: 200m
+            memory: 256Mi
+  ```  
 
 ## Non-Goals
 
 * Convert *devFlags* type to `runtime.RawExtension` to make devFlags flexible. [ref comment](https://github.com/opendatahub-io/architecture-decision-records/pull/23#issuecomment-1847519245)
+* As an incremental goal, we can introduce HPAs(only for Dashboards or services which requires auto scaling).
+  An example yaml is given below:
+  ```yaml
+    spec:
+  components:
+    dashboard:
+      managementState: Managed
+      dashboard:
+        autoscale:
+          hpa:
+            minReplicas: 1
+            maxReplicas: 10
+  ```   
+  
 ## How
 
 The idea is to create a custom plugin for the kustomize where we use JSON patch to patch these updated values and then deploy the manifests.
 
 ## Open Questions
 
-- Should we limit the max replica count to 2 in case of controllers?
-- As per current requests, only Dashboard team require HPAs. So as an initial step should we start using HPAs only for Dashboard. As for other teams, we can(maybe for now) add replica count in the devFlags and then move it to the *base* when required?
+- Should we limit the max replica count to 2 in case of controllers? (Answer)[https://github.com/opendatahub-io/architecture-decision-records/pull/23#issuecomment-1883820075]
+- As per current requests, only Dashboard team require HPAs. So as an initial step should we start using HPAs only for Dashboard. As for other teams, we can(maybe for now) add replica count in the devFlags and then move it to the *base* when required? Answer: HPA can be implemented as an incremental update.
 
 ## Alternatives
 
--- a/operator/ODH-ADR-Operator-0004-replica-image-support.md
+++ b/operator/ODH-ADR-Operator-0004-replica-image-support.md
@@ -17,7 +17,7 @@ This document explains about upgrading the DSC CRD to support customizing the re
 
 ## Why
 
-Current operator design does not support changing the replica count, images are hard-coded in the manifests and the is no way to set the resource limits/requests. A lot of teams and users have started requesting for providing support of manipulating these fields.
+Current operator design does not support changing the replica count, images are hard-coded in the manifests and there is no way to set the resource limits/requests. A lot of teams and users have started requesting for providing support of manipulating these fields.
 NOTE: The image update is targeted for dev purposes and will not be available in a supported configuration. 
 
 ## Goals
@@ -33,7 +33,21 @@ NOTE: The image update is targeted for dev purposes and will not be available in
         kfNotebookController:
             replicas: 1
   ```
-* For each component(eg. codeflare, ray, kserve etc) provide an image(s) field in the *devFlags* through which we can pass-in a custom image(s) to be used by the controller.
+* For each component(eg. codeflare, ray, kserve etc) provide an image(s) field in the *devFlags* through which we can pass-in a custom image(s) to be used by the controller. Example yaml is given below:
+  ```yaml
+    spec:
+  components:
+     workbenches:
+        managementState: Managed
+        notebookController:
+            replicas: 1
+        kfNotebookController:
+            replicas: 1
+        devFlags:
+          images:
+            odh-notebook-controller-image: quay.io/opendatahub/odh-notebook-controller:latest
+            kf-notebook-controller-image: quay.io/opendatahub/kubeflow-notebook-controller:latest
+  ```
 * Control the resource limits and requests for individual components for more fine grained control. An example yaml is given below:
   ```yaml
     spec:
--- a/operator/ODH-ADR-Operator-0004-replica-image-support.md
+++ b/operator/ODH-ADR-Operator-0004-replica-image-support.md
@@ -14,28 +14,39 @@
 ## What
 
 This document explains about upgrading the DSC CRD to support customizing the replica count, images and modifying the resource limits/requests of individual components.
+
 ## Why
 
-Current operator design doesnot support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields. 
+Current operator design doesnot support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields.
+NOTE: The image update is targeted for dev purposes and will not be available in production. 
 
 ## Goals
 
-* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count.
-* For each component(eg. codeflare, ray, kserve etc) provide an image field through which we can pass-in a custom image to be used by the controller.
-* Certain components uses other images(for eg. datascience-pipelines-controller uses `IMAGE_API_SERVER`, `IMAGES_ARTIFACT`). We need to provide support to customize these images as well.
+* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count. If a component is deploying multiple deplyoments, then something like this is preferred:
+  ```yaml
+    spec:
+  components:
+     workbenches:
+          notebookController:
+               replicaCount: <>
+          kfNotebookController:
+                replicaCount: <>
+  ```
+* For each component(eg. codeflare, ray, kserve etc) provide an image field in the *devFlags* through which we can pass-in a custom image to be used by the controller.
+* Certain components uses other images(for eg. datascience-pipelines-controller uses `IMAGE_API_SERVER`, `IMAGES_ARTIFACT`). We need to provide support to customize these images as well in the *devFlags*
 * Control the resource limits and requests for individual components for more fine grained control.
 
 ## Non-Goals
 
-* 
+* Convert *devFlags* type to `runtime.RawExtension` to make devFlags flexible. [ref comment](https://github.com/opendatahub-io/architecture-decision-records/pull/23#issuecomment-1847519245)
 ## How
 
 The idea is to create a custom plugin for the kustomize where we use JSON patch to patch these updated values and then deploy the manifests.
 
 ## Open Questions
 
-- What is the best way to implement this? Should we fetch the Deployment, read the env array, apply the changes then patch the changes.
-- The requirement suggests to update the images in env, in this case JSON patch doesnot support selecting maps in an array using the key value. It allows us to select only using the array index(here env being the array). Will the env always have the same ordering of its values? If we choose the array index method, it is pretty fragile. Either some sort of strategic merge or implementing own 'plugin'?
+- Should we limit the max replica count to 2 in case of controllers?
+- As per current requests, only Dashboard team require HPAs. So as an initial step should we start using HPAs only for Dashboard. As for other teams, we can(maybe for now) add replica count in the devFlags and then move it to the *base* when required?
 
 ## Alternatives
 
--- a/operator/ODH-ADR-Operator-0004-replica-image-support.md
+++ b/operator/ODH-ADR-Operator-0004-replica-image-support.md
@@ -17,7 +17,7 @@ This document explains about upgrading the DSC CRD to support customizing the re
 
 ## Why
 
-Current operator design does not support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields.
+Current operator design does not support changing the replica count, images are hard-coded in the manifests and the is no way to set the resource limits/requests. A lot of teams and users have started requesting for providing support of manipulating these fields.
 NOTE: The image update is targeted for dev purposes and will not be available in a supported configuration. 
 
 ## Goals
@@ -33,7 +33,7 @@ NOTE: The image update is targeted for dev purposes and will not be available in
         kfNotebookController:
             replicas: 1
   ```
-* For each component(eg. codeflare, ray, kserve etc) provide an image field in the *devFlags* through which we can pass-in a custom image to be used by the controller.
+* For each component(eg. codeflare, ray, kserve etc) provide an image(s) field in the *devFlags* through which we can pass-in a custom image(s) to be used by the controller.
 * Control the resource limits and requests for individual components for more fine grained control. An example yaml is given below:
   ```yaml
     spec:
@@ -72,9 +72,6 @@ NOTE: The image update is targeted for dev purposes and will not be available in
 The idea is to create a custom plugin for the kustomize where we use JSON patch to patch these updated values and then deploy the manifests.
 
 ## Open Questions
-
-- Should we limit the max replica count to 2 in case of controllers? (Answer)[https://github.com/opendatahub-io/architecture-decision-records/pull/23#issuecomment-1883820075]
-- As per current requests, only Dashboard team require HPAs. So as an initial step should we start using HPAs only for Dashboard. As for other teams, we can(maybe for now) add replica count in the devFlags and then move it to the *base* when required? Answer: HPA can be implemented as an incremental update.
 
 ## Alternatives
 
--- a/operator/ODH-ADR-Operator-0004-replica-image-support.md
+++ b/operator/ODH-ADR-Operator-0004-replica-image-support.md
@@ -8,7 +8,7 @@
 | Authors        | [Ajay Jaganathan](@AjayJagan), [Yauheni Kaliuta](@ykaliuta) |
 | Supersedes     | N/A |
 | Superseded by: | N/A |
-| Tickets        | [Tracker issue](https://github.com/opendatahub-io/opendatahub-operator/issues/441)|
+| Tickets        | [Tracker issue1](https://issues.redhat.com/browse/RHOAIENG-277)|
 | Other docs:    | none |
 
 ## What
@@ -28,9 +28,9 @@ NOTE: The image update is targeted for dev purposes and will not be available in
   components:
      workbenches:
         managementState: Managed
-        notebookController:
+        notebook-controller-deployment:
             replicas: 1
-        kfNotebookController:
+        odh-notebook-controller-manager:
             replicas: 1
   ```
 * For each component(eg. codeflare, ray, kserve etc) provide an image(s) field in the *devFlags* through which we can pass-in a custom image(s) to be used by the controller. Example yaml is given below:
@@ -39,9 +39,9 @@ NOTE: The image update is targeted for dev purposes and will not be available in
   components:
      workbenches:
         managementState: Managed
-        notebookController:
+        notebook-controller-deployment:
             replicas: 1
-        kfNotebookController:
+        odh-notebook-controller-manager:
             replicas: 1
         devFlags:
           images:
@@ -63,6 +63,7 @@ NOTE: The image update is targeted for dev purposes and will not be available in
             cpu: 200m
             memory: 256Mi
   ```  
+NOTE: The subcomponent names have to match exactly with the deployment names.
 
 ## Non-Goals
 
@@ -74,13 +75,13 @@ NOTE: The image update is targeted for dev purposes and will not be available in
   components:
     dashboard:
       managementState: Managed
-      dashboard:
+      odh-dashboard:
         autoscale:
           hpa:
             minReplicas: 1
             maxReplicas: 10
   ```   
-  
+  NOTE: HPA implementation is not a goal of this change.
 ## How
 
 The idea is to create a custom plugin for the kustomize where we use JSON patch to patch these updated values and then deploy the manifests.
--- a/operator/ODH-ADR-Operator-0004-replica-image-support.md
+++ b/operator/ODH-ADR-Operator-0004-replica-image-support.md
@@ -113,6 +113,7 @@ N/A
 
 ## Reviews
 
-| Reviewed by                   | Date       | Notes |
-| ----------------------------- | ---------  | ------|
-| name                          | date       | ? |
+| Reviewed by                   | Date          | Notes |
+| ----------------------------- | ---------     | ------|
+| Edson Tirelli                 | Jan 11, 2024  | ?     |
+| Andrew Ballantyne             | Jan 10, 2024  | ?     |
--- a/operator/ODH-ADR-Operator-0004-replica-image-support.md
+++ b/operator/ODH-ADR-Operator-0004-replica-image-support.md
@@ -1,0 +1,68 @@
+# Open Data Hub - Extend DataScienceCluster api to support per-Component tuning
+
+|                |            |
+| -------------- | ---------- |
+| Date           | 2023-Dec-1  |
+| Scope          | OpenDataHub operator |
+| Status         | Draft |
+| Authors        | [Ajay Jaganathan](@AjayJagan), [Yauheni Kaliuta](@ykaliuta) |
+| Supersedes     | N/A |
+| Superseded by: | N/A |
+| Tickets        | [Tracker issue](https://github.com/opendatahub-io/opendatahub-operator/issues/441)|
+| Other docs:    | none |
+
+## What
+
+This document explains about upgrading the DSC CRD to support customizing the replica count, images and modifying the resource limits/requests of individual components.
+## Why
+
+Current operator design doesnot support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields. 
+
+## Goals
+
+* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count.
+* For each component(eg. codeflare, ray, kserve etc) provide an image field through which we can pass-in a custom image to be used by the controller.
+* Certain components uses other images(for eg. datascience-pipelines-controller uses `IMAGE_API_SERVER`, `IMAGES_ARTIFACT`). We need to provide support to customize these images as well.
+* Control the resource limits and requests for individual components for more fine grained control.
+
+## Non-Goals
+
+* 
+## How
+
+The idea is to create a custom plugin for the kustomize where we use JSON patch to patch these updated values and then deploy the manifests.
+
+## Open Questions
+
+- What is the best way to implement this? Should we fetch the Deployment, read the env array, apply the changes then patch the changes.
+- The requirement suggests to update the images in env, in this case JSON patch doesnot support selecting maps in an array using the key value. It allows us to select only using the array index(here env being the array). Will the env always have the same ordering of its values? If we choose the array index method, it is pretty fragile. Either some sort of strategic merge or implementing own 'plugin'?
+
+## Alternatives
+
+N/A
+
+## Security and Privacy Considerations
+
+N/A
+
+## Risks
+
+- Providing an invalid image value can break the respective deployment.
+- Since we control the resource limits and requests, there is a possibility for over/under utilization for the resources.
+
+## Stakeholder Impacts
+
+| Group                         | Key Contacts     | Date       | Impacted? |
+| ----------------------------- | ---------------- | ---------- | --------- |
+| group or team name            | key contact name | date       | ? |
+
+
+## References
+
+* optional bulleted list
+
+## Reviews
+
+| Reviewed by                   | Date       | Notes |
+| ----------------------------- | ---------  | ------|
+| name                          | date       | ? |


operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,68 @@
+# Open Data Hub - Architecture Decision Record template
```
ykaliuta, line 1: something like `extend DataScienceCluster api to support per-Component tuning`?

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,68 @@
...
+
+This document explains about upgrading the DSC CRD to support customizing the replica count, images and modifying the resource limits/requests of individual components.
+## Why
+
+Current operator design doesnot support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields. 
+
+## Goals
+
+* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count.
+* For each component(eg. codeflare, ray, kserve etc) provide an image field through which we can pass-in a custom image to be used by the controller.
```
andrewballantyne, line 24: Would this not be a dev flag sorta option? Custom replica count makes sense to give it to the customers... custom image value seems not something "base" I would think.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,68 @@
...
+## Non-Goals
+
+* 
+## How
+
+The idea is to create a custom plugin for the kustomize where we use JSON patch to patch these updated values and then deploy the manifests.
+
+## Open Questions
+
+- What is the best way to implement this? Should we fetch the Deployment, read the env array, apply the changes then patch the changes.
```
andrewballantyne, line 37: Would configurations for HPA be better? Or is the goal to just have a fixed set of replicas? If so, is that just a step short of the long term goal or are we against dynamic scaling?

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,68 @@
...
+## Why
+
+Current operator design doesnot support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields. 
+
+## Goals
+
+* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count.
+* For each component(eg. codeflare, ray, kserve etc) provide an image field through which we can pass-in a custom image to be used by the controller.
+* Certain components uses other images(for eg. datascience-pipelines-controller uses `IMAGE_API_SERVER`, `IMAGES_ARTIFACT`). We need to provide support to customize these images as well.
+* Control the resource limits and requests for individual components for more fine grained control.
```
andrewballantyne, line 26: Curious if we could support dynamic variables -- letting each manifest approach that as needed? It would be less helpful for the users 🤔, but it would offer more flexibility.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,68 @@
...
+
+N/A
+
+## Security and Privacy Considerations
+
+N/A
+
+## Risks
+
+- Providing an invalid image value can break the respective deployment.
```
andrewballantyne, line 50: If this is a dev flag, I think it is fine... if it is a customer setting -- I'm a little concerned, but I imagine they can "undo" the effort pretty easily if the image is not pull-able.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,68 @@
...
+- Since we control the resource limits and requests, there is a possibility for over/under utilization for the resources.
```
andrewballantyne, line 51: Another reason to use HPA imo.

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: hey @andrewballantyne .. thanks for the review 👍 
so I guess I cannot talk much on the architecture level 😅 
@VaishnaviHire @etirelli  can you please help me out here

AjayJagan: here is the related issue for that: https://github.com/opendatahub-io/opendatahub-operator/issues/673

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: We didnot think for HPA, but with your idea maybe we can talk about the short term goal being the `replicas` and then we can iteratively upgrate it to HPA ... wdyt

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: Sorry to ask but what are dynamic variables 😅

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: hmm so I have a question
dev flags are for dev purposes only(if I am correct)
So in the future, what if the user wants to use his own customized image for the components?
And maybe it is the responsibility of the user to make sure that the image works.
wdyt

operator/ODH-ADR-Operator-0004-replica-image-support.md
andrewballantyne: That issue doesn't get into it, but the comments state "help for development". I'm not sure we should mix the two. 

Customers will deploy DSC resources after operator install... we should keep the API specific for that user type. 

"Overriding images" is not something I'd ever expect a customer to do, unless we explicitly walk them through it or document it. Because it has a potential to cause all kinds of problems (eg. incompatibility between images)

operator/ODH-ADR-Operator-0004-replica-image-support.md
andrewballantyne: It's fine to address it as a 2nd phase... but HPA covers replicas directly... just with some smarts of when to up/down it... I imagine the use-cases may be wide enough to want both in the long run... so maybe the simple step is best. Just voicing my opinion.

operator/ODH-ADR-Operator-0004-replica-image-support.md
andrewballantyne: I'm thinking of things like [`"<admin_groups>"`](https://github.com/red-hat-data-services/odh-manifests/blob/d72c788e8eeba638c583a8c7be2508757be1f26c/odh-dashboard/odhdashboardconfig/odhdashboardconfig.yaml#L25) -- I imagine this could be easily a variable we could use to configure things as needed for each component.

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: got it 👍 so image updates by customers can become a stability issue

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: @ykaliuta wdyt can we try HPAs

operator/ODH-ADR-Operator-0004-replica-image-support.md
andrewballantyne: If the user wants to use a specific version for a component -- I would strongly encourage us to take a hard look at what this means to the stability of the application. DevFlags allow the flexibility to "know you're creating your own unsupported path" whereas the base information is very much exposed to "hmm, I'll try this out" and it could have any number of changes that impact things. Including impacting data because of an upgrade (migrations). That makes it difficult to go back.

Custom images seems outside of a "normal flow" and that's why I'm not sure about it. (see also https://github.com/opendatahub-io/architecture-decision-records/pull/23#discussion_r1415962832)

AjayJagan:
@andrewballantyne , some questions
1. If we are going to use HPA, then should we add it as a resource along with the other kustomize manifests?(the odh-manifests repo)
2. Should this be configured for each component?
3. If HPA is configured, then users need not worry about the replica count right? So there is no need to add a custom replica functionality ? 😅 (Please correct me if I am wrong)

operator/ODH-ADR-Operator-0004-replica-image-support.md
VaishnaviHire: @andrewballantyne brings in couple of good points. What we can do is add the custom image under `devFlags` that is present under every component instead

AjayJagan: got it... so will move the custom images to `devFlags` then.

AjayJagan:
> @andrewballantyne , some questions
> 
> 1. If we are going to use HPA, then should we add it as a resource along with the other kustomize manifests?(the odh-manifests repo)
> 2. Should this be configured for each component?
> 3. If HPA is configured, then users need not worry about the replica count right? So there is no need to add a custom replica functionality? 😅 (Please correct me if I am wrong)

As @VaishnaviHire  says, can we also move the HPA configurations to `devFlags`?
Also instead of directly modifying the replica count, can we plan to modify the `minReplicas` and `maxReplicas`?
@ykaliuta ^

AjayJagan:
> > > @andrewballantyne , some questions
> > > 
> > > 1. If we are going to use HPA, then should we add it as a resource along with the other kustomize manifests?(the odh-manifests repo)
> > > 2. Should this be configured for each component?
> > > 3. If HPA is configured, then users need not worry about the replica count right? So there is no need to add a custom replica functionality? 😅 (Please correct me if I am wrong)
> > 
> > 
> > As @VaishnaviHire says, can we also move the HPA configurations to `devFlags` as well? Also instead of directly modifying the replica count, can we plan to modify the `minReplicas` and `maxReplicas`? @ykaliuta ^
> 
> I think HPA configurations/replicas can be updated by users, so they should not be included in the devFlags, just the `images`

understood 👍

andrewballantyne:
> @andrewballantyne , some questions
> 
> 1. If we are going to use HPA, then should we add it as a resource along with the other kustomize manifests?(the odh-manifests repo)
> 2. Should this be configured for each component?
> 3. If HPA is configured, then users need not worry about the replica count right? So there is no need to add a custom replica functionality ? 😅 (Please correct me if I am wrong)

So let us start at the top-level... HPA will be extra work to configure (over just Replicas). If we ever intend to support non HPA deployments (maybe the backend teams don't want HPA?) then Replicas sound like something we will need to support -- so maybe we just start there until we understand HPA edges more.

HPA is (imo) a needed feature for the Dashboard & I imagine every component (but probably used less by many). The Dashboard can get hit by 1000s of users at the same time and our pod needs to scale for that. If 10 people use it at the same time, no need to spend the resources putting up several replicas. This can always be done through pure replica's, as determined by the Installing User (eg. "I have thoughts of 1000s of users concurrently using it vs I have an intent for only 10 to be using it at any point in time" -- which would be several replicas or just one)

I'm worried at this point that HPA will be entirely dictated by me for the Dashboard needs and I think we need to avoid a single component owner driving this. Maybe this is an AC topic for a future enhancements, and we go with Replicas for now.

Thoughts @VaishnaviHire @AjayJagan ?

AjayJagan:
> > @andrewballantyne , some questions
> > 
> > 1. If we are going to use HPA, then should we add it as a resource along with the other kustomize manifests?(the odh-manifests repo)
> > 2. Should this be configured for each component?
> > 3. If HPA is configured, then users need not worry about the replica count right? So there is no need to add a custom replica functionality ? 😅 (Please correct me if I am wrong)
> 
> So let us start at the top-level... HPA will be extra work to configure (over just Replicas). If we ever intend to support non HPA deployments (maybe the backend teams don't want HPA?) then Replicas sound like something we will need to support -- so maybe we just start there until we understand HPA edges more.
> 
> HPA is (imo) a needed feature for the Dashboard & I imagine every component (but probably used less by many). The Dashboard can get hit by 1000s of users at the same time and our pod needs to scale for that. If 10 people use it at the same time, no need to spend the resources putting up several replicas. This can always be done through pure replica's, as determined by the Installing User (eg. "I have thoughts of 1000s of users concurrently using it vs I have an intent for only 10 to be using it at any point in time" -- which would be several replicas or just one)
> 
> I'm worried at this point that HPA will be entirely dictated by me for the Dashboard needs and I think we need to avoid a single component owner driving this. Maybe this is an AC topic for a future enhancements, and we go with Replicas for now.
> 
> Thoughts @VaishnaviHire @AjayJagan ?
> 
> EDIT: Let me answer your questions in case you guys do want to go ahead with HPA.
> 
> 1. The Operator will definitely need to entirely manage this, no implementation should come from any component, this is outside of their world
> 2. Yes
> 3. That's my hope -- but like I noted in the block above, does everyone want to be HPA-enabled?

Looking at it from your perspective, we may need to draw a line between frontend and backend apps(for using HPA) so if you ask me let's do it iteratively, so the first version of this can be replicas and maybe we only provide HPA configurability for the dashboard(or if other teams require this, we can extend the HPA functionality) -- this is just my thoughts. Any new idea is welcome 😄 . And thanks for the detailed explanation

andrewballantyne:
@jwforres Perhaps this is just the inefficiency of the Dashboard pod 🤔 No backend component ever needs more pods if they have direct API calls in the 1000s concurrently? I may be off base here then.

Replicas for the Dashboard at least are 5 by default for RHOAI; to support Perf requests. ODH is 2 by default. Are we the only RHOAI deployment that needs to be controlled based on the number of users?

AjayJagan:
So if I understand this correct
Front-end = configurable replica count(maybe not it the devFlags) - In future we can consider HPAs
Back-end = configurable replica count only in the devFlags(maybe we can have a limit set there to max = 2)
Is this assumption right?
@jwforres @andrewballantyne @VaishnaviHire ^

jwforres:
I clarified in slack but will do so here as well - my comment about only 2 replicas was specific to controllers. Backend API servers may still require scaling beyond 2.  With controllers only 1 replica is ever serving as the active one, so there is no benefit to going beyond having a single passive backup. That same situation isn't true with backend API servers that are serving requests.

AjayJagan:
> I clarified in slack but will do so here as well - my comment about only 2 replicas was specific to controllers. Backend API servers may still require scaling beyond 2. With controllers only 1 replica is ever serving as the active one, so there is no benefit to going beyond having a single passive backup. That same situation isn't true with backend API servers that are serving requests.

ok, now it's clear - `the why`... Thanks for the clarification 😄

bartoszmajsak:
I understand that the main driver for this change is giving dev teams the ability to slim down deployments so they can be run in the constrained clusters. If that is the case I think this would be a great opportunity to rework `devFlags` and make them flexible. What I was thinking of is to make `devFlag` a `*runtime.RawExtension` field so that each component can model settings fitting their needs by providing their own structs matching exactly all the resources they wish to make configurable. 

From a broader perspective, incorporating elements like replicas directly into the DSC API appears to be a somewhat leaky abstraction. This approach could lead to further additions of similar nature in the future, complicating the API. Additionally, the meaning of a parameter like `replicas: 2` becomes ambiguous, particularly for components responsible for creating multiple Deployments (aside from controllers).

I think providing some reasonable defaults through means of `HorizontalPodAutoscaler` and perhaps `PodDisruptionBudget` and letting users fine-tune them might be a better option to start with.

vaibhavjainwiz:
I am very much agree with @bartoszmajsak suggestion to expose `*runtime.RawExtension` which could be easily customize to take differet input. 

I am seeing this requirement as starting of various customization which could be required to expose for different component. Currently we are maintaining single CRD `DataScienceCluster` where we are trying to capture every customization of every component. Soon it would became big and nightmare to maintain.

**Proposal**
In addition to @bartoszmajsak suggestion, I am proposing to maintain seperate CRDs for each ODH component where every component team design them as per their need.
like Seperate CRD for:
- DashboardConfiguration
- NotebookConfiguration
- ModelServingConfiguration

For example:
In ModelMesh Serving, there is no single image which user could just override. KServe component use to run 4 container in a pod so user could provide customize image for any of it. 
I am proposing to maintain something like below:

**ModelServingConfiguration.yaml**
```
apiVersion: opendatahub.io/v1
kind: ModelServingConfiguration
metadata:
  name: default
spec:
  odh-model-controller:
    replica: 1
    image: quay.io/vajain/odh-model-controller:2.0
    manifests:
      contextDir: config
      sourcePath: ''
      uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'
  kserve:
    replica: 2
    kserve-agent
      image: quay.io/vajain/kserve-agent:2.0
    kserve-storage-initializer
      image: quay.io/vajain/kserve-storage-initializer:2.0
```

I know its completely contradictory to current design but if we follow above described approach then we could complete remove the `DataScienceCluster` CRD. To enable and disable specific component we could rely on Component CRD like if user create an instance of  `ModelServingConfiguration` CRD then he want to enable ModelServing in the project. Its very similiar how `Serveless` team provide support for `knative-serving` and `knative-eventing`

@danielezonca

vaibhavjainwiz:
Any progress on this one?

AjayJagan:
@VaishnaviHire @andrewballantyne I have updated the ADR a bit. Let me know how it looks :)

AjayJagan:
> Any progress on this one?

Hi @vaibhavjainwiz AFAIK, the idea of separate CRDs can be put into a separate ADR 😅 . wdyt?

operator/ODH-ADR-Operator-0004-replica-image-support.md
strangiato: I would say that setting replica counts is a must have and HPA is a nice to have.  Replica count should be much easier to impliment but if you design it correctly it shouldn't be too much work to extend it to support an HPA as a phase 2.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Stakeholder Impacts
+
+| Group                         | Key Contacts     | Date       | Impacted? |
+| ----------------------------- | ---------------- | ---------- | --------- |
+| group or team name            | key contact name | date       | ? |
```
andrewballantyne, line 68: Dashboard team approves.

HPA discussions were a bit of derail, apologizes for that. The simple Replica value is fine for me.

andrewballantyne (approved):

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,79 @@
...
+| Tickets        | [Tracker issue](https://github.com/opendatahub-io/opendatahub-operator/issues/441)|
+| Other docs:    | none |
+
+## What
+
+This document explains about upgrading the DSC CRD to support customizing the replica count, images and modifying the resource limits/requests of individual components.
+
+## Why
+
+Current operator design doesnot support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields.
```
strangiato, line 20: `doesnot` needs a space between the two words.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,79 @@
...
+NOTE: The image update is targeted for dev purposes and will not be available in production.
```
strangiato, line 21: `in production` should be `in a supported configuration` or something along those lines.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,79 @@
...
+This document explains about upgrading the DSC CRD to support customizing the replica count, images and modifying the resource limits/requests of individual components.
+
+## Why
+
+Current operator design doesnot support changing the replica count and the images are hard-coded in the manifests. A lot of teams and users have started requesting for providing support of manipulating these fields.
+NOTE: The image update is targeted for dev purposes and will not be available in production. 
+
+## Goals
+
+* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count. If a component is deploying multiple deplyoments, then something like this is preferred:
```
strangiato, line 25: I would recommend following this same pattern even for components that do create a single deployment.  For example, the Dashboard only deploys the dashboard deployment, but maybe down the line the Dashboard team has a need to create a second separate deployment for a new feature.  By starting with this as the default, you won't have to break the API later on to add configurations for that new deployment.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,79 @@
...
+    spec:
+  components:
+     workbenches:
+          notebookController:
+               replicaCount: <>
+          kfNotebookController:
+                replicaCount: <>
+  ```
+* For each component(eg. codeflare, ray, kserve etc) provide an image field in the *devFlags* through which we can pass-in a custom image to be used by the controller.
+* Certain components uses other images(for eg. datascience-pipelines-controller uses `IMAGE_API_SERVER`, `IMAGES_ARTIFACT`). We need to provide support to customize these images as well in the *devFlags*
```
strangiato, line 36: These should not be set in the DSC.  These should be set in the DataSciencePipelineApplication object which is actually deploying those images.  I believe these options are already supported there as well.

strangiato:
I'm catching up on this thread but wanted to weigh in on a couple of different things.

I'm 100% in alignment with @andrewballantyne on focusing on setting replicas for the dashboard and focusing on HPA afterwards.  Replicas should be a much easier change and HPA's are a nice to have.  I would recommend considering how HPAs might be integrated into that at a later phase.

Like @andrewballantyne said, having the dashboard scaled to 5 replicas is great for 1000 active users but most customers may have 10-15 users at any given time.  On top of that I would say that 95% of the time users are working with OpenShift-AI they are working on things outside of the Dashboard (such as in a notebook and the OCP console) and only 5% of the time they are actively working in the Dashboard.

As for setting images, that was originally suggested before DevFlags existed.  I don't expect it should be a "supported" feature for customers.  I think that including it under the DevFlags is perfectly fine.  I think that if customers want to deploy their own image, we probably need to better understand why they need their own image (e.g. adding custom certs to the image) and see about building that capability into the operator as a first class feature instead of just using their own image.  

I have previously voiced some concerns around the DevFlag but I will say them again here.  DevFlags seem like a bit of a kludgy way to say "this is supported and this isn't supported" instead of just documenting it.  This creates a problems like maybe we do decide that custom images are supported for components down the road.  Do we need to change the API spec to make that policy change if it is currently located under devFlags?

For the controllers I will make the argument that they should NEVER be more than a single replica.  Having a standby replica is a waste of resources and IMO should have never been introduced.  Controllers are designed around eventual consistency and there is no guarantee of "response time" from a controller.  So what is the risk of a controller crashing with a single pod?  The resource you are attempting to deploy takes 45 seconds instead of 30 seconds to deploy?  The controller going down doesn't cause the resource it has deployed to stop working.  So your model server API endpoint still functions if the modelmesh-controller goes down, it just won't get updated if the resource definition is modified until the pod comes back up.

So in my ideal solution all of the controllers should be 1 replica by default and there is no need to even add the replicas feature outside of the dashboard.  If a component teams insists that they need an HA deployment of the controller, I need to have the ability to set it to 1 replica for Single Node Deployment or one where we have some resource sensitivity. 

Regarding the `ModelServingConfiguration` idea...   What value does having this as a separate object have over doing it in the existing DSC?  It is still a one to one relationship between DSC and ModelServingConfiguration so it appears to me to just be moving the configuration from one object to another.  I would prefer to keep the configuration of these types of items consistent across all components and keep them in a single location.  If we want ant move to a unique object per component we should do away with the DSC entirely and only have unique objects for each component.  Mixing the two is just going to cause confusion on "what should I set where?".

My ideal configuration of the DSC would look something like this:

```
apiVersion: datasciencecluster.opendatahub.io/v1
kind: DataScienceCluster
metadata:
  name: default
spec:
  components:
    codeflare:
      managementState: Removed
    dashboard:
      managementState: Managed
      dashboard: <1>
        autoscale: <2>
          enabled: true
          hpa:
            minReplicas: 1
            maxReplicas: 10
            metrics:
              - type: Resource
                resource:
                name: cpu
                target:
                  type: Utilization
                  averageUtilization: 50
    datasciencepipelines:
      managementState: Managed
      data-science-pipelines-operator-controller-manager:
        resources:
          requests:
            cpu: 100m
            memory: 128Mi
          limits:
            cpu: 200m
            memory: 256Mi
    kserve:
      managementState: Removed
    modelmeshserving:
      managementState: Managed
      odh-model-controller: <3>
        replicas: 1
        image: quay.io/vajain/odh-model-controller:2.0 <4>
      modelmesh-controller:
        replicas: 1
      devFlags:
        manifests:
          contextDir: config
          sourcePath: ''
          uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'
    ray:
      managementState: Removed
    workbenches:
      managementState: Managed
```

1 - The dashboard component today only deploys a single set of pods but by setting these options under `spec.dashboard.dashboard` we give ourselves more flexibility if the Dashboard chooses to add additional pods in the future as part of the Dashboard component.
2 - This autoscaling configuration is based on how ArgoCD handles setting up Autoscaling for it's components.  I know RH has an aversion to "true/false" values in the CRD so that could be changed or maybe even just the presence of the autoscaling/hpa sections is enough to infer that we want it enabled.  HPA could potentially be moved up to `spec.dashboard.dashboard.hpa` but I think keeping it under autoscaling probably gives us more flexibility in the long run.  For example the controllers may benefit from having an option to setup VPA in the future.
3 - The modelmeshserving component is a good example of something that deploys multiple sets of pods with `odh-model-controller` and `modelmesh-controller`.  We need to be able to control these independently so having the pattern of `spec.<component>.<deployment>` like with the dashboard makes a lot of sense here.
4 - Including the image here was always what I imagined before the devFlag idea was introduced.  We could set this same value under `spec.modelmeshserving.devFlags.odh-model-controller.image` as well but it seems like a lot of redundant structure to the object.  I don't think anyone will question that we aren't going to support a custom image that they built themselves though.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,79 @@
...
+## Non-Goals
+
+* Convert *devFlags* type to `runtime.RawExtension` to make devFlags flexible. [ref comment](https://github.com/opendatahub-io/architecture-decision-records/pull/23#issuecomment-1847519245)
+## How
+
+The idea is to create a custom plugin for the kustomize where we use JSON patch to patch these updated values and then deploy the manifests.
+
+## Open Questions
+
+- Should we limit the max replica count to 2 in case of controllers?
```
strangiato, line 48: I mentioned it in my larger comment already, but I'll mention it here as well.  Controllers should never have more than 1 replica.

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: okay got it 👍

vaibhavjainwiz:
@strangiato 
> If we want ant move to a unique object per component we should do away with the DSC entirely and only have unique objects for each component. Mixing the two is just going to cause confusion on "what should I set where?".

Same thought came to me as well to completly remove the DSC and ony define independent object for each component but  earlier restrict my self because it would be a big architectural change. But I belive this is the cleaner way of doing it. WDYT?

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: I have a question. Shouldn't we provide a way to configure these images as well? I was thinking of a scenario where a user might to test these images as well, and through DSC we pass these images as env variables to the DSP `data-science-pipelines-operator-controller-manager`? Please correct me if I am wrong :)

operator/ODH-ADR-Operator-0004-replica-image-support.md
etirelli: +1 to what andrew said. It is the user's responsibility to ensure the images are correct if he replaces them by setting the devflag parameter for development purposes. We do not support replacing images in production though.

operator/ODH-ADR-Operator-0004-replica-image-support.md
etirelli: Shouldn't the choice to support HPA be done per component? My understanding is that the operator can't assume every component will have proper metrics/probes implemented to support HPA. The component teams need to ensure that first... or maybe I am missing something.

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,107 @@
...
+
+## References
+
+* optional bulleted list
+
+## Reviews
+
+| Reviewed by                   | Date       | Notes |
+| ----------------------------- | ---------  | ------|
+| name                          | date       | ? |
```
etirelli, line 107: lgtm

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,104 @@
...
+| Tickets        | [Tracker issue](https://github.com/opendatahub-io/opendatahub-operator/issues/441)|
+| Other docs:    | none |
+
+## What
+
+This document explains about upgrading the DSC CRD to support customizing the replica count, images and modifying the resource limits/requests of individual components.
+
+## Why
+
+Current operator design does not support changing the replica count, images are hard-coded in the manifests and the is no way to set the resource limits/requests. A lot of teams and users have started requesting for providing support of manipulating these fields.
```
ykaliuta, line 20: `the is` -> `there is` I guess

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Goals
+
+* For each component(eg. codeflare, ray, kserve etc) provide a field `replicas` through which we can modify the the Deployment's replica count. An example yaml is given below:
+  ```yaml
+    spec:
+  components:
+     workbenches:
+        managementState: Managed
+        notebookController:
```
ykaliuta, line 31: I would prefer these names not in a free from but exactly corresponding the object's `Name` value. Like you did below. Probably also mention it in the doc. It will make filtering much easier.

Also could it be reasonable to have it as an array of some sort of `deployments` then? It will make handling of it on the struct Component level much more simple. Or what do you think?

zdtsw:
> modelmeshserving:
>       managementState: Managed
>       odh-model-controller: <3>
>         replicas: 1
>         image: quay.io/vajain/odh-model-controller:2.0 <4>
>       modelmesh-controller:
>         replicas: 1
>       devFlags:
>         manifests:
>           contextDir: config
>           sourcePath: ''
>           uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'

I have been thinking about this part:
```
modelmeshserving:
      managementState: Managed
      odh-model-controller: <3>
        replicas: 1
        image: quay.io/vajain/odh-model-controller:2.0 <4>
      modelmesh-controller:
        replicas: 1
      devFlags:
        manifests:
          contextDir: config
          sourcePath: ''
          uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'
```
if user prefer using devFlags, why would they set 2 deployments` odh-model-controller` and `modelmesh-controller`?
it would be much easier to config whatever they want(image, replica, resources etc) in the remote tarball in one go.

so the question is,  if the difficulty of using devFlags becomes a hinder which starts to introduce other options, then maybe we should not even keep devFlags, instead of having "<component>.deployment.<deploymentX>" only (deploymentX will be provided in an official way so the users know what exactly to put there)
something like: 
```
modelmeshserving:
      managementState: Managed
      deployments:
        odh-model-controller:     // need to set to exact name to match whats been deployed in the cluster, or it wont work
           image: quay.io/my/image1:tag
           resources:
              requests:
                 cpu: 100m
                 memory: 128Mi
              limits:
                 cpu: 200m
                memory: 256Mi
        modelmesh-controller:         // if the deploymentX does not contains "controller" then replica value will be valid
           image: quay.io/my/image2:tag
           replica:  2         // here will fall back to 1 since it is a controller
           .....
       
```

AjayJagan:
> I have been thinking about this part:
> 
> ```
> modelmeshserving:
>       managementState: Managed
>       odh-model-controller: <3>
>         replicas: 1
>         image: quay.io/vajain/odh-model-controller:2.0 <4>
>       modelmesh-controller:
>         replicas: 1
>       devFlags:
>         manifests:
>           contextDir: config
>           sourcePath: ''
>           uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'
> ```
> 
> if user prefer using devFlags, why would they set 2 deployments` odh-model-controller` and `modelmesh-controller`? it would be much easier to config whatever they want(image, replica, resources etc) in the remote tarball in one go.
> 
> so the question is, if the difficulty of using devFlags becomes a hinder which starts to introduce other options, then maybe we should not even keep devFlags, instead of having ".deployment." only (deploymentX will be provided in an official way so the users know what exactly to put there) something like:
> 
> ```
> modelmeshserving:
>       managementState: Managed
>       deployments:
>         odh-model-controller:     // need to set to exact name to match whats been deployed in the cluster, or it wont work
>            image: quay.io/my/image1:tag
>            resources:
>               requests:
>                  cpu: 100m
>                  memory: 128Mi
>               limits:
>                  cpu: 200m
>                 memory: 256Mi
>         modelmesh-controller:         // if the deploymentX does not contains "controller" then replica value will be valid
>            image: quay.io/my/image2:tag
>            replica:  2         // here will fall back to 1 since it is a controller
>            .....
>        
> ```

@andrewballantyne @etirelli  thoughts on this?

andrewballantyne:
We need the tarball manifest uri -- this is key to overriding everything. The `image` is a QoL as an often use-case is "I want everything standard, but I want my PR's custom image, or some feature built image ... so you have to fork and make a new manifest file. This is more hinderance to nightly images and things that are quickly updated than it is for pinned releases and updates. 

I agree with Trevor, we should probably not look to bloat it up high -- but as I noted in the AC, devFlags have a lot of value and removing them is not a great idea imo. If we don't want to expand it, it is a flexible mechanism for it not to be a blocker going forward.

etirelli:
@AjayJagan I agree with Trevor's points, except for having the image outside of devflags. The purpose of devflags is to allow developers like us to have the flexibility of overriding settings during the development cycle, without confusing users about what they should or should not use. Overriding an image is not something I expect users should ever do.

zdtsw:
> @andrewballantyne @etirelli thanks for clarifying again. So we stick to the devFlags idea and may be iteratively incorporate Trevor's idea(apart from the images)... wdyt @zdtsw ?

If I understand @AjayJagan you correctly:
2 valid configurations for devFlags ? 
(for example 1)

```
modelmeshserving:
      managementState: Managed
      devFlags:
        odh-model-controller:     // need to set to exact name to match whats been deployed in the cluster, or it wont work
           image: quay.io/my/image1:tag
           resources:
              requests:
                 cpu: 100m
                 memory: 128Mi
              limits:
                 cpu: 200m
                memory: 256Mi
        modelmesh-controller:         // if the deploymentX does not contains "controller" then replica value will be valid
           image: quay.io/my/image2:tag
           replica:  2         // here will fall back to 1 since it is a controller
           .....
```
  and example 2 (as what we are having already) :
```
modelmeshserving:
      managementState: Managed
      devFlags:
        manifests:
          contextDir: config
          sourcePath: ''
          uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'
```

AjayJagan:
> > @andrewballantyne @etirelli thanks for clarifying again. So we stick to the devFlags idea and may be iteratively incorporate Trevor's idea(apart from the images)... wdyt @zdtsw ?
> 
> If I understand @AjayJagan you correctly: 2 valid configurations for devFlags ? (for example 1)
> 
> ```
> modelmeshserving:
>       managementState: Managed
>       devFlags:
>         odh-model-controller:     // need to set to exact name to match whats been deployed in the cluster, or it wont work
>            image: quay.io/my/image1:tag
>            resources:
>               requests:
>                  cpu: 100m
>                  memory: 128Mi
>               limits:
>                  cpu: 200m
>                 memory: 256Mi
>         modelmesh-controller:         // if the deploymentX does not contains "controller" then replica value will be valid
>            image: quay.io/my/image2:tag
>            replica:  2         // here will fall back to 1 since it is a controller
>            .....
> ```
> 
> and example 2 (as what we are having already) :
> 
> ```
> modelmeshserving:
>       managementState: Managed
>       devFlags:
>         manifests:
>           contextDir: config
>           sourcePath: ''
>           uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'
> ```

So the `resources` and `replicas` are planned to be moved out of devFlags. Only the `image` field will remain as part of the devFlags. 
And I think there should be 2 configurations 
1 - The manifest URI
2 - The image(direct update)

vaibhavjainwiz:
Could we please create a seperate ADR to followup on images part

vaibhavjainwiz:
In our last **OpenShift AI - Architecture Council** meeting we had some consensus that maintaining separate CRDs for each component would be a viable way to tackle current challenges. So to avoid polluting the current ADR I had started the discussion thread on slack to raise this point.
https://redhat-internal.slack.com/archives/C05KDB2HFQQ/p1705989784479989

bartoszmajsak:
It would be great to use e.g. GitHub discussions instead of internal channels if we want to keep this open to the community.

operator/ODH-ADR-Operator-0004-replica-image-support.md
LaVLaS: I agree that HPA should be implemented in the controller since it assumes the application is designed to scale horizontally.   IMO, this dev feature should only support the ability to modify the existing numbers for non-production use cases

operator/ODH-ADR-Operator-0004-replica-image-support.md
```diff
@@ -0,0 +1,119 @@
...
+            memory: 128Mi
+          limits:
+            cpu: 200m
+            memory: 256Mi
+  ```  
+
+## Non-Goals
+
+* Convert *devFlags* type to `runtime.RawExtension` to make devFlags flexible. [ref comment](https://github.com/opendatahub-io/architecture-decision-records/pull/23#issuecomment-1847519245)
+* As an incremental goal, we can introduce HPAs(only for Dashboards or services which requires auto scaling).
```
LaVLaS, line 70: Is this ADR feature required in order to support HPA in supported configurations?

operator/ODH-ADR-Operator-0004-replica-image-support.md
LaVLaS: I agree that this should have strict matching for the Deployment & Container names since they should be a unique pair across the Data Science Cluster.

AjayJagan: you mean something like 
`notebook-controller-deployment`, `odh-notebook-controller-manager` is this correct?

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: well it was discussed to support replicas for now and then maybe as a next step we can do the HPA(we need to analyze which components need HPAs). So I added it as a non goal for this ADR 😅 . Maybe I can remove the wordings like, `incremental goal` if needed :)

ykaliuta: Do I understand right that HPA does not require any operator interaction and can be easily deployed by Component's manifests?

operator/ODH-ADR-Operator-0004-replica-image-support.md
ykaliuta: If we need to implement it in the operator how to distinguish different controller types?

At the moment it looks for me that components should configure HPA themselves (they know better their requirements) without additional load on the user. For the rare cases to force scaling down replicas to 1 devFlag  can be used (wondering why HPA could be not enough)

But I can still miss the whole picture.

operator/ODH-ADR-Operator-0004-replica-image-support.md
AjayJagan: maybe we can provide the min/max replicas from our DSC?

zdtsw: Let's leave HPA out of the scope of this ADR if it is not immediately applied to what we want to achieve here.
and yes to @ykaliuta component can opt-in if they want to utilize HPA by creating yaml in their manifests on top of their deployment.

AjayJagan: agreed

operator/ODH-ADR-Operator-0004-replica-image-support.md
LaVLaS: From an offline conversation, the discussion was to ensure that the object names mentioned in the DSC for replica & image overrides match the `name` fields from `kustomize build <COMPONENT MANIFEST>` to prevent any confusion and a 1:1 mapping to  `.metadata.name -> .spec.containers[].name -> (.replaces | .image)`

AjayJagan:
@LaVLaS I have updated the ADR with the changes haveing 1:1 mapping between the subcomponent names & the deployment. Please let me know if this looks fine :)

VaishnaviHire:
Splitting this ADR to get more focussed reviews for configuring `resources` - https://github.com/opendatahub-io/architecture-decision-records/pull/32

PR #25, opened 2024-01-24
ADR-0005: Introduce a controller for Data Science Projects

VaishnaviHire:
- This ADR proposes introduction of Namespace controller in ODH Operator
- This change is essential to include cert bundles in every Data Science Project

### DIFF
--- a/operator/ODH-ADR-Operator-0005-namespace-controller.md
+++ b/operator/ODH-ADR-Operator-0005-namespace-controller.md
@@ -1,0 +1,60 @@
+# Introducing namespace controller to manage Data Science Projects
+
+
+|                |                                                 |
+| -------------- |-------------------------------------------------|
+| Date           | 2024-01-24                                      |
+| Scope          |                                                 |
+| Status         | Draft                                           |
+| Authors        | [Vaishnavi Hire](@VaishnaviHire)                |
+| Supersedes     | N/A                                             |
+| Superseded by: | N/A                                             |
+| Tickets        | https://issues.redhat.com/browse/RHOAIENG-2186  |
+| Other docs:    | none                                            |
+
+## What
+
+This document outlines the decision to introduce a namespace controller in the ODH Operator.
+
+## Why
+
+Currently, the ODH Operator has no knowledge of the Data Science Projects(namespaces) managed by the ODH Dashboard.
+Introduce a controller to manage Data Science Projects through backend. This will allow components to access DSProjects 
+and add artifacts to project namespaces directly without creating then through the Dashboard.
+
+## Goals
+
+* ODH operator initialize and manage any resources that are required for ODH components.  This could include configmaps,
+secrets, namespaced resources.
+* ODH operator would be responsible for reconciling these resources to ensure that they are consistent with each release.
+
+
+## Non-Goals
+
+* This implementation does not include introduce a new CR, and will just watch Namespace resource.
+
+## How
+
+* Introduce a new controller in the operator that watches for namespaces identified as Data Science Projects.
+
+## Alternatives
+
+* Use ODH Dashboard to deploy resources in every Data Science Project required by ODH components. 
+
+
+## Stakeholder Impacts
+
+| Group     | Key Contacts     | Date       | Impacted? |
+|-----------| ---------------- | ---------- | --------- |
+| Platform  | key contact name | date       | ? |
+| Dashboard | key contact name | date       | ? |
+
+## References
+
+* optional bulleted list
+
+## Reviews
+
+| Reviewed by                   | Date       | Notes |
+| ----------------------------- | ---------  | ------|
+| name                          | date       | ? |


VaishnaviHire:
cc: @zdtsw @LaVLaS @andrewballantyne

zdtsw:
does this mean, we will move the logic out of dashboard and into the new namespace controller instead?

bartoszmajsak:
> This change is essential to include cert bundles in every Data Science Project

Is the introduction of something like https://cert-manager.io/ on the roadmap?

VaishnaviHire:
> https://cert-manager.io/

Not yet, https://issues.redhat.com/browse/RHOAIENG-1926 describes the scope of the feature.


Regarding the project controller - 

> Interestingly enough I was talking with @aslakknutsen about moving some bits of https://github.com/maistra/odh-project-controller/ to the opendatahub-operator.
> 
> We created this controller to support featured-flagged service mesh in DSPs, so it takes care of
> 
> * enrolling namespace created through dashboard to service mesh
> * add authz rules
> * exposes certain settings through configmaps so they can be reused across other controllers
> 
> I am happy to port this code instead of maintaining it as a separate one, but I wonder if we are on the same page with the scope here.

Yes I see this as one of the usecase for the controller. For initial implementation, idea is to be able to identify which namespaces are set as Data Science Projects and add custom configmap to it. We can extend the controller to include above configuration for Servicemesh.

VaishnaviHire:
> does this mean, we will move the logic out of dashboard and into the new namespace controller instead?

Eventually, but the immediate goal for this controller would be ensuring we are able to identify namespaces set as Data Science Project and are able to manage and deploy resources to these namespaces

operator/ODH-ADR-Operator-0005-namespace-controller.md
```diff
@@ -0,0 +1,60 @@
...
+| Tickets        | https://issues.redhat.com/browse/RHOAIENG-2186  |
+| Other docs:    | none                                            |
+
+## What
+
+This document outlines the decision to introduce a namespace controller in the ODH Operator.
+
+## Why
+
+Currently, the ODH Operator has no knowledge of the Data Science Projects(namespaces) managed by the ODH Dashboard.
```
strangiato, line 21: It isn't clear what the goal is here.

Adding objects to a project is a technical implementation, but it isn't clear what you want to achieve by adding these objects.

Why do you want to add secrets or config maps to a data science project?  What problems does this solve?

operator/ODH-ADR-Operator-0005-namespace-controller.md
```diff
@@ -0,0 +1,60 @@
...
+
+* This implementation does not include introduce a new CR, and will just watch Namespace resource.
+
+## How
+
+* Introduce a new controller in the operator that watches for namespaces identified as Data Science Projects.
+
+## Alternatives
+
+* Use ODH Dashboard to deploy resources in every Data Science Project required by ODH components.
```
strangiato, line 42: I think it is important to distinguish between "A user deploys objects using the dashboard" and "the dashboard automatically deploys things".

Obviously a user making a choice to deploy specific objects is totally fine.

However, the dashboard should be focused on acting as a thin UI and should contain minimal logic.  It should not do things like "user wants to create a project, so I am also going to create these other objects".

strangiato (commented):
There is very little detail in the why here.  Especially for what impact it has on the end users.

When I create a new project today, I'm not aware of any secrets or config maps that need to be created for a project to function correctly, so what is this controller going to solve?

Another consideration, is that in the dashboard now allows you to work in projects that do not have the ODH label to make them a data science project.  How will those projects be handled? 

My suspicion is that whatever you want to do should probably be handled by the existing controllers.  For example, if you want to setup a secret or config map for Data Science Pipelines, that secret or config map should really just be created by the DSPA instead.

If you want to create something that is shared by multiple controllers, we need to better understand the dependencies between those controllers.  For example, if the model registry is going to be dependent on DSPA, the model registry should be able to look at the DSPA and get the info it needs from it or from objects the DSPA manages.  We should not create a new controller that both are now dependent on.

operator/ODH-ADR-Operator-0005-namespace-controller.md
LaVLaS: The goal is to provide a mechnanism for the operator to manage DS-Project where that logic is currently embedded in the Dashboard application code requiring the Dashboard application to own the logic of DS-Project lifecycle.  The first pass of this is to support the ability to manage RHOAI global resources across all DS-Project namespaces so that we can mantain functionality in environments where the Dashboard is not installed

Immediate concerns this will address are self-signed certificates that need to be made available across all DS-Project namespaces.  If we standardized DS-Project control then we can have an initialization and management phase for common resource that are global to all components.  We don't envision this to be a heavy controller that replaces the logic that is best handled by a single component controller.

operator/ODH-ADR-Operator-0005-namespace-controller.md
LaVLaS: > However, the dashboard should be focused on acting as a thin UI and should contain minimal logic.
This is the overwhelming consensus across the teams including the Dashboard team

> It should not do things like "user wants to create a project, so I am also going to create these other objects".
There are no plans to change the current UX from user choice of what is installed and running.  There may be future ODH shared resources that are required as "ODH infrastructure dependencies" within a DS-Project that we want to make visible to the operator

operator/ODH-ADR-Operator-0005-namespace-controller.md
andrewballantyne, line 21: It is a misnomer to be considering "DSProjects" as a "managed by the ODH Dashboard" -- as I noted in the cert meeting, we don't manage anything. We don't watch anything. We render, we create existing known CRD items,  and we update existing K8s objects. In some cases, we have services we deploy and we "manage" that in some fashion... but I think there is some ambiguity to who really manages those (DSP & Trusty CRs).

Projects not created in the Dashboard do show up in the Dashboard, and due to some unfortunate needs (Service Mesh, Model Mesh, KServe) we need to modify the project to apply labels and annotations.

Today we continue to use the `opendatahub.io/dashboard` label on all projects we create, but since we show both, it seemed harmless. That is not guaranteed to be the future of the Dashboard nor is it intended to be something others look for.

I'm a little sensitive to the wording, but if basically all I said is already in agreement (which sounds like it is by other comments) then please ignore me.

operator/ODH-ADR-Operator-0005-namespace-controller.md
```diff
@@ -0,0 +1,60 @@
...
+
+This document outlines the decision to introduce a namespace controller in the ODH Operator.
+
+## Why
+
+Currently, the ODH Operator has no knowledge of the Data Science Projects(namespaces) managed by the ODH Dashboard.
+Introduce a controller to manage Data Science Projects through backend. This will allow components to access DSProjects 
+and add artifacts to project namespaces directly without creating then through the Dashboard.
+
+## Goals
```
andrewballantyne, line 25: Maybe this is not an ADR thing -- but how are you folks going to find the projects that are considered "Data Science Projects"?

LaVLaS:
> There is very little detail in the why here. Especially for what impact it has on the end users.

The goal is not to impact the user by extending the lifecycle support to ODH infrastructure deployed in DS-Projects

> When I create a new project today, I'm not aware of any secrets or config maps that need to be created for a project to function correctly, so what is this controller going to solve?

This is the reason for controller because right now the Dashboard application is responsible for creating any required secrets or configmaps for a DS project to function correctly

> Another consideration, is that in the dashboard now allows you to work in projects that do not have the ODH label to make them a data science project. How will those projects be handled?

There is no intent to restrict the view to DS-Projects with ODH labels but if ODH has a presence in those DS Projects we need a mechanism to watch and manage any core ODH infrastructure in those projects

> My suspicion is that whatever you want to do should probably be handled by the existing controllers. For example, if you want to setup a secret or config map for Data Science Pipelines, that secret or config map should really just be created by the DSPA instead.

Agreed. This does not replace the setup that is owned by an individual component

> If you want to create something that is shared by multiple controllers, we need to better understand the dependencies between those controllers.  For example, if the model registry is going to be dependent on DSPA, the model registry should be able to look at the DSPA and get the info it needs from it or from objects the DSPA manages. We should not create a new controller that both are now dependent on.

This is a good point but this DS-Project would only be responsible for a shared dependency when ODH infrastructure is the owner of the dependency, like a shared Database or global ODH configuration.

operator/ODH-ADR-Operator-0005-namespace-controller.md
bartoszmajsak: > Projects not created in the Dashboard do show up in the Dashboard, and due to some unfortunate needs (Service Mesh, Model Mesh, KServe) we need to modify the project to apply labels and annotations.

Not quite sure why Service Mesh is called out in the first place, as it's still behind the feature flag :)  But as I've been implictly called out I thought I will share my $.2.

These "unfortunate" needs (I believe what you mean here is putting extra annotations on a "project" for mentioned components?) call for shifting this responsibility to a controller(s) instead. 

Then the only thing the user would need to know is how to make their namespace part of Open Data Hub if it is not created through the Dashboard. Maybe this is a good opportunity to think about this now. By looking at how Istio does I think we might be good with a label. This will be picked up by controller(s) participating in Open Data Hub and... the rest of the "magic" will follow.

bartoszmajsak: Could we create a new, standalone project for managing self-signed certificates? This might help keep our operators' tasks and codebases clear and focused.

operator/ODH-ADR-Operator-0005-namespace-controller.md
bartoszmajsak: I opened another PR, which is not an ADR yet, but more a RFC (Request For Comments) at this point - hence in draft.

#26

LaVLaS: Since we are not proposing a new CR then it would be a `label` for namespaces where ODH components are present

strangiato:
There is lots of great comments and detail in this discussion thread about what the goals of introducing a new controller are and I hope these make it back into the ADR.

> This is the reason for controller because right now the Dashboard application is responsible for creating any required secrets or configmaps for a DS project to function correctly

What secrets and ConfigMaps is the Dashboard responsible for creating today?  I'm not aware of anything special that the dashboard creates when creating a project but I haven't dug around to check.  Providing concrete examples of those kinds of things on the ADR would be very helpful.

> if ODH has a presence in those DS Projects we need a mechanism to watch and manage any core ODH infrastructure in those projects

Is "core ODH infrastructure" the controllers that are deployed in `opendatahub` today?  Or are you talking about CR's managed by the controllers?

What are you want to watch that can't be watched already?  The Notebook Controller can see Notebook and the child objects it creates.  DSP can see DPSA's and the different resources it manages?  What does this proposed controller need to see that it can't be seen by other controllers already?

> Since we are not proposing a new CR then it would be a label for namespaces where ODH components are present

Is this going to be the existing label the Dashboard uses or a new label?

What happens if the proposed controller does see a project that has an ODH related object in it that doesn't have the label already?  Will the controller add the label automatically?

What happens when the label is added to a project or a new project is created with the label?  Does the controller need to create something in that namespace?  What is the purpose of that the thing that it creates?

> shared Database 

Sharing a database is an anti-pattern that should be avoided.

> global ODH configuration

I would love to know more about what kinds of global configurations you are thinking about here.  Presumably these would be things that need to be applied into each data science project somehow?  What would prevent these from living in the `opendatahub` namespace?

LaVLaS:
> What happens if the proposed controller does see a project that has an ODH related object in it that doesn't have the label already? Will the controller add the label automatically?
> What happens when the label is added to a project or a new project is created with the label? Does the controller need to create something in that namespace? What is the purpose of that the thing that it creates?

All good questions which we'll update in the ADR. 

> Is "core ODH infrastructure" the controllers that are deployed in opendatahub today? Or are you talking about CR's managed by the controllers?

Core ODH infrastructure is a general reference to common configs or objects that affect the successful deployment or integration between ODH components within a DS-Project namespace.  As of right now, the only thing planned is self-signed certificate bundles that need to be replicated across DS-Project namespaces.   

> What does this proposed controller need to see that it can't be seen by other controllers already?

This is a mechanism to provide any ODH global configurations that may need to be in place within DS-Projects before any dependant components are deployed to a namespace.  

> What are you want to watch that can't be watched already?

DS-Projects namespaces are not watched already

> Sharing a database is an anti-pattern that should be avoided.

It is an example of a shared service in a DS-Project that two or more components may need but neither one owns the lifecycle.  Presently, there is no ODH entity, outside of the Dashboard, that can ensure shared services or  namespace configurations are applied correctly

> I would love to know more about what kinds of global configurations you are thinking about here. Presumably these would be things that need to be applied into each data science project somehow? What would prevent these from living in the opendatahub namespace?

DS-Projects are for namespaced objects that are restricted to a namespace.  That example would assume that a DS-Project user or service has access to the `opendatahub` namespace

operator/ODH-ADR-Operator-0005-namespace-controller.md
VaishnaviHire: My understanding was by looking for the `opendatahub.io/dashboard` label, we should be able to identify the projects.

operator/ODH-ADR-Operator-0005-namespace-controller.md
israel-hdez: > Could we create a new, standalone project for managing self-signed certificates? 

Related to this, cert-manager org has a project named _trust-manager_: https://github.com/cert-manager/trust-manager

If distributing trust is all what we need for self-signed certs support, me may use it. Although, not sure if it is really available to us.

operator/ODH-ADR-Operator-0005-namespace-controller.md
```diff
@@ -0,0 +1,60 @@
...
+
+* ODH operator initialize and manage any resources that are required for ODH components.  This could include configmaps,
+secrets, namespaced resources.
+* ODH operator would be responsible for reconciling these resources to ensure that they are consistent with each release.
```
israel-hdez, line 29: This is confusing.

Would the new controller be bundled-in odh-operator? or would it be a separate `Deployment`?

operator/ODH-ADR-Operator-0005-namespace-controller.md
```diff
@@ -0,0 +1,60 @@
...
+| Status         | Draft                                           |
+| Authors        | [Vaishnavi Hire](@VaishnaviHire)                |
+| Supersedes     | N/A                                             |
+| Superseded by: | N/A                                             |
+| Tickets        | https://issues.redhat.com/browse/RHOAIENG-2186  |
+| Other docs:    | none                                            |
+
+## What
+
+This document outlines the decision to introduce a namespace controller in the ODH Operator.
```
israel-hdez, line 17: Looks like this ADR would result in scaffolding a new controller, but is going to be useless initially. So, as it stands I think it is weak.

By looking the other comments, this would be used as implementation detail to support self-signed certificates. So, I wonder if it isn't better to propose the new controller as part of a concrete need? With this I mean that it may be better to replace this ADR with another one for the self-signed certs support, and mention the new controller as part of the implementation.

operator/ODH-ADR-Operator-0005-namespace-controller.md
LaVLaS: This would be a controller in the operator.

operator/ODH-ADR-Operator-0005-namespace-controller.md
andrewballantyne: For posterity, we talked about this and the label is not enough anymore. If anyone wants to know more, there is a [slack thread](https://redhat-internal.slack.com/archives/C05KDB2HFQQ/p1706108405416829) about the back and forth of it.

operator/ODH-ADR-Operator-0005-namespace-controller.md
andrewballantyne: > Not quite sure why Service Mesh is called out in the first place, as it's still behind the feature flag :) 

This is irrelevant fwiw... you're part of the architecture as it stands... not being available to RHOAI or having/not having a feature flag is irrelevant and your use case needs to be counted for in all scenarios that talk about the manipulation of Projects.

---

More to your response -- I think the entire set of components needs to collectively get a better idea of how one interacts with the Project. I've gotten comments about the over-usage of labels and annotations on things like Projects, because they are not normal objects... you cannot modify these without touching the Namespace object -- which requires special circumventing of the RBAC of the user to do (our service account does this). CRs is the preferred way as I have been informed, as they are within the realm of user control & can be listed across namespaces by controllers.

I'm not pushing this, personally I don't overly care today, but there are some requirements on upstream (Model Mesh) that prevent a full shift today -- but I'll put a topic on the Architectural Council to bring this up again and see about moving away from this heavy Project-modification methodology we have -- as I would like to remove all our custom flows related to this.

bartoszmajsak: > This is irrelevant fwiw

@andrewballantyne sure, it was supposed to be an icebreaker. But I am afraid the joke fell flat :)

I think labelling and DSP CRD are two orthogonal, yet equally important, aspects that should be discussed. Having CR is a great idea and will let us model our domain better but it also opens up more granular access control as you hinted. It also opens up a possibility for having DSPs spanning multiple namespaces.

> I would like to remove all our custom flows related to this.

bartoszmajsak: > This is irrelevant fwiw

@andrewballantyne sure, it was supposed to be an icebreaker. But I am afraid the joke fell flat :)

I think labelling and DSP CRD are two orthogonal, yet equally important, aspects that should be discussed in detail. Having CR is a great idea and will let us model our domain better but it also opens up more granular control as you hinted.

> I would like to remove all our custom flows related to this.

+ 100 on this

PR #27, opened 2024-01-25
docs: Proposal to provide better customization flexibility at compone…

vaibhavjainwiz:
…nt level

This ADR proposes the approach to provide better customization flexibility to all components of Open Data Hub.

## Merge criteria:

### DIFF
--- a/operator/ODH-ADR-Operator-0005-seperate-crd.md
+++ b/operator/ODH-ADR-Operator-0005-seperate-crd.md
@@ -35,7 +35,7 @@ Replicas field in Dashboard component makes complete sense but if user defines r
 
 ## Non-Goals
 
-- Retroactively update existing DataScienceCluster immediately. This will be a gradual process.
+- Defining the configuration/parameters of each component 
 - Define application/runtime specific configuration.
 
 ## How
--- a/operator/ODH-ADR-Operator-0005-seperate-crd.md
+++ b/operator/ODH-ADR-Operator-0005-seperate-crd.md
@@ -64,21 +64,21 @@ kind: ModelServingConfiguration
 metadata:
  name: default
 spec:
- odh-model-controller:
- replica: 1
- image: quay.io/vajain/odh-model-controller:2.0
- manifests:
- contextDir: config
- sourcePath: ''
- uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'
- kserve:
- replica: 2
- managementState: Managed
- memoryLimit: 1Gi,
- cpuRequest: 100m,
- cpuLimit: 1
- modelmesh:
- managementState: Managed 
+	odh-model-controller:
+		replica: 1
+		image: quay.io/vajain/odh-model-controller:2.0
+		manifests:
+			contextDir: config
+			sourcePath: ''
+			uri: 'https://github.com/opendatahub-io/odh-model-controller/tarball/main'
+	kserve:
+		replica: 2
+		managementState: Managed
+		memoryLimit: 1Gi,
+		cpuRequest: 100m,
+		cpuLimit: 1
+	modelmesh:
+		managementState: Managed 
 ```
 Every component would have its own CRD and reconciler.
 
--- a/operator/ODH-ADR-Operator-0005-seperate-crd.md
+++ b/operator/ODH-ADR-Operator-0005-seperate-crd.md
@@ -1,0 +1,105 @@
+# Open Data Hub - ODH component Integration with DataScienceCluster
+
+|                |                         |
+| -------------- |-------------------------|
+| Date           | 2024-01-25              |
+| Scope          | OpenDataHub operator    |
+| Status         | Draft                   |
+| Authors        | [Vaibhav Jain](@vajain) |
+| Supersedes     | N/A                     |
+| Superseded by: | N/A                     |
+| Tickets        |                         |
+| Other docs:    | none                    |
+
+## What
+This ADR proposes the approach to provide better customization flexibility to all components of Open Data Hub.
+
+## Why
+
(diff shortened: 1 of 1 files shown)


operator/ODH-ADR-Operator-0005-seperate-crd.md
```diff
@@ -0,0 +1,105 @@
...
+- **Increased Component Customization:** To provide customization flexibility at component level so that every component could define their own structure to capture user inputs.
+- **Improved fault isolation:** If one component reconciler encounters a fault or failure because of customization, it doesn’t propagate across the entire system.
+- **Quicker deployment time:** If a user made configuration changes in a specific component then only that component would reconcile.
+- **Enhance team productivity:** Allow individual teams to concentrate on their component deployment and maintenance without being burdened by the complexities of the entire system.
+- **Accelerate scalability:** New components could be onboard without causing any impact on existing functionality.
+
+## Non-Goals
+
+- Retroactively update existing DataScienceCluster immediately. This will be a gradual process.
+- Define application/runtime specific configuration.
```
HumairAK, line 39: I'm not sure how I feel about this one. When the ODH operator deploys all of the components manifests, it is also deploying any configmaps that specify run time configurations for said component's operator. While the ODH operator need not be aware of the contents of this run time configuration, it does own it yes? So any changes made to said configmap can be overwritten by the ODH operator? How does one change these run time configurations then if not through the ODH Operator which owns these?

israel-hdez: They way I understand it, is that defining the configuration/parameters of each component isn't a goal of this ADR. The goal is just to provide flexibility.

To accomplish that goal, this ADR is proposing to split the `DataScienceCluster` to several CRDs. I guess the component owners would need to create their own ADRs exposing what configs are desired to expose.

I'm not sure if we would have a tangible outcome if this ADR is accepted. The way it is written seems to be just to discuss and agree working towards a goal, but it does not propose anything specific; i.e. all shown YAML are more like examples than a hard proposal.

vaibhavjainwiz: yes @israel-hdez that the goal of with this ADR. Once we decide the approach we want to opt, we could work on the independent ADR for each component to define their customization requirement.

vaibhavjainwiz:
@etirelli @danielezonca Could you please review

strangiato (commented):
I have to state that I am vehemently opposed to the idea of breaking the DSC up into multiple objects.

I firmly believe that breaking apart the DSC into multiple sub components does not solve the problems described here will simply make the issue of managing/installing ODH more challenging for it's end users.

The original intention of the DSC was to simplify the installation process for end users, and allow the operator to have the logic to know how to install different components together, as opposed to acting as a bad gitops controller that simply attempts to blindly install manifests.

If the DSC is not able to install components in the correct combination, that is an issue with the logic in the controller, not an issue with the DSC, which is simply an interface.

I do think there is a need to evolve the DSC and the structure of the spec, but nothing that is proposed here could not be accomplished within the DSC.  In fact, the proposed structure seems like it is actually going back towards the very first draft of the DSC I created:

https://github.com/strangiato/odh-operator-api/blob/main/api-examples/DataScienceCluster/DataScienceCluster.yaml 

Breaking the DSC up into multiple objects is going to create unnecessary complexity in the installation process.  Having multiple objects now means that users need to create 5+ objects to use all of the features of ODH as opposed to creating a single object that contains all of the logic for installing the components.

Part of the goal of the operator redesign was that we wanted to simplify the install as much as possible.  Having a single required object makes it so admins are prompted to create that single object.  We are able to present them with a sane, default recommended setup, that includes as many features as possible, to allow them to explore the most capabilities.  When we require only specific features, we make it easy to pick and choose which components they wish to turn off and the operator is responsible for making that happen.

Now, let's talk about why you would break this into multiple objects....

1) You require the ability to have separate RBAC for those objects.  You want to certain users to be able to create objects of one kind, and other users to create other objects.  For example, I may allow admins to create ArgoCD instances, but want my developers to create application objects, without the ability to create an ArgoCD instance.  Having separate objects allows us to have separate RBAC for who can do what.

Does this apply to the DSC: No. The expectation is that a single person with cluster admin is installing and configuring all of the components.

2) The objects have a one to many relationship between each other.  Good examples of this include things like Strimzi/AMQStreams, with a Kafka cluster, and Topics.  You can have many topics per Kafka instance so you need to have separate objects to handle that many to one relationship.  You could handle this as a list, but more on that in the next reason why you would break up the object.

Does this apply to the DSC: No.  ODH Components are expected to only have one instance and one instance only of all components.

3) The objects have very different lifecycles from each other and users are expected to interact with them at different phases of a project.  Going back to the Kafka example, I am likely to setup my Kafka cluster initially, and I may create new topics during many different phases of the development process.  I don't want to constantly be updating the kafka object just to create new topics and have to dig through the more complex object/accidentally mess up the cluster while doing that.

Does this apply to the DSC: No.  The expectation is that the components are configured at the same time and edits are not regular.  You may which to change some of the things that are installed at some point, but you would likely be evaluating/making decisions that require you to have the larger context in mind at that time so having all of the component configurations in a single place is a benefit.

4) Keeping all of the logic of the configurations in a single object is too much context to manage in a single location.  The object is simply too big to effectively be able to manage it.

Does this apply to the DSC: No. The DSC is growing in complexity, but it is not overly complex.  The majority of "options" we are discussing should not be visible to users unless they actively choose to set them.  Things like the Manifests should never be visible to an end user or populated by default.  Options like setting images or replicas should be handled in a similar fashion.  90% of users will not ever need these options or ever even need to know that they could set them.  A simple set of managed/removed flags should be sufficient for the majority of use cases.

PR #28, opened 2024-02-12
 Make Trusted Bundle Configmap available to components

VaishnaviHire:
## Merge criteria:

### DIFF
--- a/operator/ODH-ADR-0004-odh-trusted-ca-configmap.md
+++ b/operator/ODH-ADR-0004-odh-trusted-ca-configmap.md
@@ -1,0 +1,69 @@
+# Open Data Hub - Make Trusted Bundle Configmap available 
+
+
+|                |                                                             |
+| -------------- |-------------------------------------------------------------|
+| Date           | 2024-02-12                                                  |
+| Scope          | Open Data Hub                                               |
+| Status         | Draft                                                       |
+| Authors        | [Landon LaSmith](@LaVLaS), [Vaishnavi Hire](@VaishnaviHire) |
+| Supersedes     | N/A                                                         |
+| Superseded by: | N/A                                                         |
+| Tickets        |                                                             |
+| Other docs:    | none                                                        |
+
+## What
+
+Add trusted-cabundle configmap to all non-openshift namespaces on ODH operator installation.
+
+## Why
+
+The first step to support self-signed certificates in ODH deployments is to make trusted-cabundle available in all ODH namespaces.
+
+This allows ODH components to mount the certs as part of their deployment VolumeMounts.
+
+## Goals
+
+* Make trusted-cabundle configmap available in all non-openshift namespaces
+* Users can opt-out of configmap injection by explicitly adding `security.opendatahub.io/inject-trusted-ca-bundle=false` annotation to a given namespace.
+
+## Non-Goals
+
+* Modification or management of OpenShift-specific or default namespaces concerning the trusted-cabundle configmap.
+* Removal of injected trusted-cabundle configmap
+
+## How
+
+* We are introducing a controller that will be responsible for creating trusted-cabundle configmap in all new and existing non-openshift namespaces.
+* For trusted-cabundle configmap, we are standardizing on `odh-trusted-ca-bundle` as the configmap name with a label of `app.kubernetes.io/part-of=opendatahub-operator`.
+* A namespace is considered non-openshift if -
+  * It doesn't start with `openshift-`
+  * It doesn't start with `kube-`
+  * It is not `openshift`
+  * It is not `default`
+* The configmap injection is triggered using an api field in DSCI `.spec.trustedCABundle.managementState`. When set to a `Managed` state this will inject the cert configmap in
+all non-openshift namespaces. Users can opt-out of cert injection by setting the managementState to `Removed`.
+
+## Alternatives
+
+1. We are proposing a longer term solution that involves adding cert-cabundle configmap to only the namespaces that have
+ODH resources. This approach is contingent upon the successful implementation of [DataScienceProjects controller ADR](https://github.com/opendatahub-io/architecture-decision-records/pull/25).
+
+
+## Stakeholder Impacts
+
+| Group                 | Key Contacts      | Date       | Impacted? |
+|-----------------------|-------------------| ---------- | --------- |
+| ODH Dashboard Team    | @andrewballantyne | date       | ? |
+| IDE Team              | @harshad16        | date       | ? |
+| DS Pipelines Team     | @HumairAK         | date       | ? |
+| Serving Team          | @Jooho            | date       | ? |
+| TrustyAI Team         | @RobGeada         | date       | ? |
+| Docs Team             | Manuela Ansaldo   | date       | ? |
+| Distributed Workloads | @anishasthana     | date       | ? |
+
+## Reviews
+
+| Reviewed by   | Date       | Notes |
+|---------------|------------| ------|
+| Edson Tirelli | 2024-02-12 | ? |


ODH-ADR-0006-odh-trusted-ca-configmap.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Goals
+
+* Make trusted-cabundle configmap available in all non-openshift namespaces
+* Users can opt-out of configmap injection by explicitly adding `security.opendatahub.io/inject-trusted-ca-bundle=false` annotation to a given namespace.
+
+## Non-Goals
+
+* Modification or management of OpenShift-specific or default namespaces concerning the trusted-cabundle configmap.
+
```
etirelli, line 33: Maybe add something to the effect of:
- removal of injected trusted-cabundles
- injection of trusted-cabundles for namespaces that were previously skipped due to the annotation `security.opendatahub.io/inject-trusted-ca-bundle=false`

ODH-ADR-0006-odh-trusted-ca-configmap.md
```diff
@@ -0,0 +1,61 @@
...
+## How
+
+* We are introducing a controller that will be responsible for creating trusted-cabundle configmap in a new and existing non-openshift namespaces.
```
etirelli, line 36: "...in all new..."

ODH-ADR-0006-odh-trusted-ca-configmap.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. We are proposing a longer term solution that involves adding cert-cabundle configmap to only the namespaces that have
+ODH resources. This approach is contingent upon the successful implementation of [DataScienceProjects controller ADR](https://github.com/opendatahub-io/architecture-decision-records/pull/25).
+
+
+## Stakeholder Impacts
+
+| Group                         | Key Contacts     | Date       | Impacted? |
+| ----------------------------- | ---------------- | ---------- | --------- |
+| group or team name            | key contact name | date       | ? |
```
etirelli, line 54: add components and also the docs team?

VaishnaviHire: Yeah, will update this

ODH-ADR-0006-odh-trusted-ca-configmap.md
VaishnaviHire: > injection of trusted-cabundles for namespaces that were previously skipped due to the annotation security.opendatahub.io/inject-trusted-ca-bundle=false

Controller will add the configmap to the namespace when the annotation is removed

etirelli (approved):
lgtm

ODH-ADR-0006-odh-trusted-ca-configmap.md
```diff
@@ -0,0 +1,67 @@
...
+* Users can opt-out of configmap injection by explicitly adding `security.opendatahub.io/inject-trusted-ca-bundle=false` annotation to a given namespace.
+
+## Non-Goals
+
+* Modification or management of OpenShift-specific or default namespaces concerning the trusted-cabundle configmap.
+* Removal of injected trusted-cabundle configmap
+
+## How
+
+* We are introducing a controller that will be responsible for creating trusted-cabundle configmap in all new and existing non-openshift namespaces.
```
HumairAK, line 37: will the configmap name/labels be as  landon has described [here](https://issues.redhat.com/browse/RHOAISTRAT-60?focusedId=24058309&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24058309)?

> We are standardizing on **odh-trusted-ca-bundle** with a label of **app.kubernetes.io/part-of=opendatahub-operator**

If so I think we should mention that here.

Also what is the recommendation for components when looking for the existence of the configmap?  

Do we search on the label? configmap name? both? I'd assume it's just the label, but I think we should be explicit here so all components look up logic is relatively consistent.

VaishnaviHire: Yes I will update to include the configmap description.

I think it will be better to `search by name`, reason being `app.kubernetes.io/part-of=opendatahub-operator` is a generic label and can retrieve duplicate resources in future

ODH-ADR-0006-odh-trusted-ca-configmap.md
```diff
@@ -0,0 +1,68 @@
...
+| Serving Team          | @Jooho            | date       | ? |
+| TrustyAI Team         | @RobGeada         | date       | ? |
+| Docs Team             | Manuela Ansaldo   | date       | ? |
+| Distributed Workloads | @anishasthana     | date       | ? |
+
+## Reviews
+
+| Reviewed by   | Date       | Notes |
+|---------------|------------| ------|
+| Edson Tirelli | 2024-02-12 | ? |
```
HumairAK, line 68: /lgtm

lucferbux (approved):
/lgtm

VaishnaviHire:
Should this be merged @etirelli ?

harshad16 (commented):
/lgtm

thanks.

LaVLaS:
@VaishnaviHire This has enough approvals so we can merge this.  The only change necessary is to relocate it to the [operator](./operator) folder now that #29 has been merged

ODH-ADR-0006-odh-trusted-ca-configmap.md
```diff
@@ -0,0 +1,68 @@
...
+## Why
+
+The first step to support self-signed certificates in ODH deployments is to make trusted-cabundle available in all ODH namespaces.
+
+This allows ODH components to mount the certs as part of their deployment VolumeMounts.
+
+## Goals
+
+* Make trusted-cabundle configmap available in all non-openshift namespaces
+* Users can opt-out of configmap injection by explicitly adding `security.opendatahub.io/inject-trusted-ca-bundle=false` annotation to a given namespace.
```
bbrowning, line 28: Is this pattern used elsewhere in ODH already where we assume the cluster is somewhat dedicated to ODH and do something in every non-system namespace, allowing users to opt specific namespaces out of changes? Or will this be a new change imposed on ODH users, where they might be surprised at this behavior or consider it undesirable?

gregsheremeta: > [do] we assume the cluster is somewhat dedicated to ODH

This is a great question. I work on the team and I would also like to know the answer :)

bbrowning: I'm not part of the team and can't answer you definitively @gregsheremeta , but when I've lead teams working on operators in the past we always had to assume there were other workloads and/or other operators installed in the cluster outside of our control. And, as a general rule of thumb, we wouldn't touch anything in the cluster except the workloads we're directly responsible for operating.

I don't have the full backstory here, but from looking at #25 it seems like the team generally knows that this solution of adding a configmap to every namespace in the cluster is at most a temporary workaround while a long-term solution is found. You don't know what tools may be managing objects in other namespaces in the cluster, and it would be bad to cause competing operator hotloops for example as some other tool keeps removing an unexpected configmap in a namespace it manages while opendatahub keeps adding it back.

This is just my opinion, not to be treated as anything more than that. It could be that this is the best option available for now, and I don't have the full context available to understand or make that decision on your behalf.

LaVLaS: > Is this pattern used elsewhere in ODH already where we assume the cluster is somewhat dedicated to ODH and do something in every non-system namespace, allowing users to opt specific namespaces out of changes? 

We have never assumed a cluster or "user" namespace is dedicated to ODH but there is an expectation that every (non-system) namespace in the cluster could have some component of ODH present (Workbench, Model Server, DSPA, ...).

> Or will this be a new change imposed on ODH users, where they might be surprised at this behavior or consider it undesirable?

There are scenarios where this could be considered undesirable which is why it can be disable per namespace or global.  In cases where `it would be bad to cause competing operator hotloops` it can be disabled for individual namespaces with a label above OR globally by setting the `.spec.trustedCABundle.managementState = Removed` in `DSCInitialization`.

bbrowning: > We have never assumed a cluster or "user" namespace is dedicated to ODH but there is an expectation that every (non-system) namespace in the cluster could have some component of ODH present (Workbench, Model Server, DSPA, ...).

Right, I get this and it sounds like the underlying issue is that you don't yet have a way today to identify only those namespaces that actually have ODH components present, thus the need to assume every namespace might.


> There are scenarios where this could be considered undesirable which is why it can be disable per namespace or global. In cases where `it would be bad to cause competing operator hotloops` it can be disabled for individual namespaces with a label above OR globally by setting the `.spec.trustedCABundle.managementState = Removed` in `DSCInitialization`.

I agree, it's good to have options to disable. I assume there was some discussion about defaulting the CA Bundle injection on vs off and the tradeoffs there considered? I'm not trying to change your mind here at all, but I am asking the questions that came to my mind as I read the ADR and try to understand the context that led to the decision.

If you feel it's relevant, some of the information already provided in this thread might be useful to include in the ADR itself. At least for me, it sheds helpful light on both the forces at play here and the larger context around the decision.

ODH-ADR-0006-odh-trusted-ca-configmap.md
```diff
@@ -0,0 +1,68 @@
...
+| Status         | Draft                                                       |
+| Authors        | [Landon LaSmith](@LaVLaS), [Vaishnavi Hire](@VaishnaviHire) |
+| Supersedes     | N/A                                                         |
+| Superseded by: | N/A                                                         |
+| Tickets        |                                                             |
+| Other docs:    | none                                                        |
+
+## What
+
+Add trusted-cabundle configmap to all non-openshift namespaces on ODH operator installation.
```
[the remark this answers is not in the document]
LaVLaS: We do performance and scale testing (thousands of concurrent users) in our downstream releases so any issues uncovered there or in the upstream community would escalate any planned discussions on how to improve the management of self-signed certificates across ODH user namespaces

See the Performance and Scale blog post on [A Guide to Scaling OpenShift Data Science to Hundreds of Users and Notebooks](https://www.redhat.com/en/blog/a-guide-to-scaling-openshift-data-science-to-hundreds-of-users-and-notebooks) as an early reference for the type of testing performed during our product release cycle

ODH-ADR-0006-odh-trusted-ca-configmap.md
VaishnaviHire: I will update the ADR with the information from the thread.
Note we are enabling the `trustedCABundle` feature only on new install, not on upgrade.  

One of the considerations to enable it by default in DSCI, was that the cert configmap is injected as part of the initial configuration step. The components that utilize these certs are deployed later as part of another CR(DSC). Defaulting to enabled, we can avoid the initial restart of the component deployments.

LaVLaS (approved):
/lgtm

Thx.  This covers the current functionality that is present in the v2.8 release

PR #45, opened 2024-06-03
ODH-ADR-MS-0001: KServe Private Endpoint

Jooho:
## Description

Submit ADR to change kserve(knative/istio) network topology for the private endpoint of kserve.

## Merge criteria:


model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,89 @@
...
+- The new Istio ingress gateway would forward requests to knative-local-gateway
+- i.e. we build on top of Knative routing needs (or we chain KServe routing needs with Knative ones)
+- The ISTIO_MUTUAL configuration on knative-local-gateway is no longer an issue.
+- Since both KServe and Knative gateways belong to the mesh, they can use mTLS.
+- KServe gateway would be capable of listening either to plain HTTP, or have a TLS-simple configuration.
+
+![proposed architecture](./images/ODH-ADR-MS-0001-kserve-private-network-in-cluster-img-1.png)  
+
+## Open Questions
+Does UI need to add a URL for private?
```
lucferbux, line 44: In modelmesh we display internal routes for the service, they are based on the isvc service and it displays both the grpc and http ports.
Screenshot 2024-06-03 at 17 42 12
Screenshot 2024-06-03 at 17 41 13

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,89 @@
...
+| Status         | Review                                                           |
+| Authors        | [Jooho Lee](@JOOHO)                                              |
+| Supersedes     | N/A                                                              |
+| Superseded by: | N/A                                                              |
+| Tickets        | [Tracking Issue](https://issues.redhat.com/browse/RHOAIENG-7918) |
+| Other docs:    | none                                                             |
+
+
+## What
+This ADR addresses the issue with Kserve Serverless using Istio and mTLS, which requires client certificates between pods. Applications accessing the isvc endpoint via service hostname in the same cluster must include an Istio sidecar, or they will fail.
```
ReToCode, line 16: You might want to add that http instead of mTLS would solve this issue, but is not possible because of the requirement to always have encrypted communications.

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Why
+Customers deploying ODH/RHOAI on OpenShift Cluster do not expect to include their services in the mesh to access deployed models. From their perspective, models deployed through KServe are just another service, and changing the network layer for access is burdensome. Therefore, we need to provide an out-of-the-box solution allowing access to inference services via internal hostnames, similar to other services.
+
+## Goals
+- Pods without an Istio sidecar should have access to the isvc endpoint.
+- Ensure a seamless transition for users from ModelMash to KServe.
```
ReToCode, line 23: ModelMash -> ModelMesh

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,89 @@
...
+
+## How
+It involves adding an Ingress Gateway and integrating with OpenShift Certificates services.
+The new Istio ingress gateway would only be used for internal requests.
+- KServe would no longer share the knative-local-gateway, with regards to configurations.
+- The new Istio ingress gateway would forward requests to knative-local-gateway
+- i.e. we build on top of Knative routing needs (or we chain KServe routing needs with Knative ones)
+- The ISTIO_MUTUAL configuration on knative-local-gateway is no longer an issue.
+- Since both KServe and Knative gateways belong to the mesh, they can use mTLS.
+- KServe gateway would be capable of listening either to plain HTTP, or have a TLS-simple configuration.
```
ReToCode, line 39: "listening either to plain HTTP" can be removed IMHO, as this could also just be done on knative-local-gateway, and would not make sense to do it in an additional layer.

ReToCode (commented):
minor things inline, generally what we discussed; 

approved

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,89 @@
...
+
+![proposed architecture](./images/ODH-ADR-MS-0001-kserve-private-network-in-cluster-img-1.png)  
+
+## Open Questions
+Does UI need to add a URL for private?
+Does Knative/istio team support the proposed way?
+Can Service Mesh team make the solution more efficient?
```
danielezonca, line 46: Nitpicking: replace Service Mesh with Istio or replace Knative/Istio with Serverless/Service Mesh in the previous question.
Service Mesh is product name while Knative/Istio are community names. I'm fine with both but we should be consistent.

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,89 @@
...
+
+![proposed architecture](./images/ODH-ADR-MS-0001-kserve-private-network-in-cluster-img-1.png)  
+
+## Open Questions
+Does UI need to add a URL for private?
+Does Knative/istio team support the proposed way?
+Can Service Mesh team make the solution more efficient? 
+
+## Alternatives
+Knative team members designed the solution below for consideration. The benefit of the solution being in Knative layer is sharing tests coverage within the component and have lighter the ODH/KServe layer on top of it. This alternative solution is not available in Knative yet and may take a time to be done, so it can be considerable in the future once available. In contrast, the main solution proposed in this document is for the short term in ODH releases.
```
danielezonca, line 49: It also requires cert-manager that has larger set of implications for the whole ODH project

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
danielezonca, line 16: I think the What should be extended to describe the issue: the title of the ADR is about having "private endpoints" while here the sentence is about Istio/mTLS etc.
I would mention something like that by default endpoints are exposed and encrypted by Istio but users might want to keep the model endpoint private.

danielezonca (approved):
Minor comments, overall I'm fine with the proposal.

Approved

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: For KServe, it is not possible to expose two ports simultaneously, making it different from ModelMesh. This issue is related to the epic [RHOAIENG-5222](https://issues.redhat.com/browse/RHOAIENG-5222). The main idea is to have the external endpoint disabled by default. If users want to expose the external endpoint, they can enable it as needed.
So the URL should be internal URL and external URL. Both URL can be used for grpc and rest based on the isvc configuration. From my 2 cents, we don't need to specify rest or grpc for kserve. @israel-hdez @spolti  @Xaenalt WDYT?

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: correct, do you want me to add this to ADR?

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: I agree with you. We should use TLS only.

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: Yes, I agree, but this is not a public/private issue; it is about communication issues within the cluster. The current network architecture only allows requests that are under the mesh. With this change, we aim to handle requests that are outside the mesh as well. So why don't we change the title like `KServe Handling Requests Outside the Mesh` ?

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: updated

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
spolti: @Jooho what do you mean with `The main idea is to have the external endpoint disabled by default. If users want to expose the external endpoint, they can enable it as needed.`?

so, when a user creates a model, if the want to use it, they will need to manually expose it?

not sure if I followed what is being discussed.
The Jiras in question, are not clear to me as well, I am probably missing the broader context.

Jooho: > so, when a user creates a model, if the want to use it, they will need to manually expose it?
correct. That's what BU team wanted. 
So default behavior -> No route. Only internal communication is possible
if the user wants -> Route can be created by annotation.

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: updated

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
ReToCode: I think so, its not really about mTLS, more about having https instead of http IMHO.

Jooho: This `What` part explains the current situation. So I think most of the comment is right but let me add this part at the end.
~~~
This ADR addresses the issue with Kserve Serverless using Istio and mTLS, which requires client certificates between pods. Applications accessing the isvc endpoint via service hostname in the same cluster must include an Istio sidecar, or they will fail. Without adding an Istio sidecar, it is currently impossible to access the isvc endpoint in the cluster.
~~~
Could you please review it ?

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: updated

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
israel-hdez: > From my 2 cents, we don't need to specify rest or grpc for kserve. @israel-hdez @spolti @Xaenalt WDYT?

It shouldn't be needed, because KServe provides only a single endpoint.

Sadly, since there is no support to expose more than one port, usually in the ServingRuntime you need to choose between using GRPC or REST. But this is at runtime definition. For the shown endpoints, there is consistency regardless of the protocol.

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,83 @@
...
+| Status         | Review                                                           |
+| Authors        | [Jooho Lee](@JOOHO)                                              |
+| Supersedes     | N/A                                                              |
+| Superseded by: | N/A                                                              |
+| Tickets        | [Tracking Issue](https://issues.redhat.com/browse/RHOAIENG-7918) |
+| Other docs:    | none                                                             |
+
+
+## What
+git
```
ReToCode, line 16: Probably not intended change?

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: yes I agreed. WDYT ? @lucferbux

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: oops my bad. rollabcked.

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
lucferbux: Mmmm so PM should be aware that we are not gonna match here feature parity, since modelmesh will allow to display different urls, if this is on par to what we have right now for public routes I'm ok.
And yes, by default the **checkbox** will be disabled, we are gonna mimic the same behavior we have in modelmesh, but all the current deployed models are still gonna be public in a migration right @Jooho 

As far as PM is ok with us not having multiple ports I'm ok with it and I can approve this.

Jooho: It seems that these issues arise because KServe and ModelMesh are fundamentally different. Therefore, this page should have different sections for each component.

lucferbux: Ok, I'm moving this conversation to Slack to ping Adam, but I can pre-approve this.

lucferbux (approved):
/lgtm

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,83 @@
...
+
+| Reviewed by           | Date        | Approval | Notes |
+| --------------------- | ----------- | -------- | ----- |
+| Jooho Lee             | Jun 3, 2024 | Approved |       |
+| Edgar Hernandez       | Jun 3, 2024 | Approved |       |
+| Alessandro Lazarotti  | Jun 3, 2024 | Approved |       |
+| Daniele Zonca         | Jun 3, 2024 | Approved |       |
+| Reto Lehmann          | Jun 3, 2024 | Approved |       |
+| Rob Cernich           |             |          |       |
+| Lucas Fernadez aragon |             |          |       |
```
lucferbux, line 81:
```suggestion
| Lucas Fernadez aragon | Jun 12, 2024| Approved |       |
```

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,83 @@
...
+- The ISTIO_MUTUAL configuration on knative-local-gateway is no longer an issue.
+- Since both KServe and Knative gateways belong to the mesh, they can use mTLS.
+- KServe gateway would be capable of a TLS-simple configuration.
+
+![proposed architecture](./images/ODH-ADR-MS-0001-kserve-private-network-in-cluster-img-1.png)  
+
+## Open Questions
+Does UI need to add a URL for private?
+Does Serverless/ServiceMesh team support the proposed way?
+Can ServiceMesh team make the solution more efficient?
```
bartoszmajsak, line 46: I am not sure I understand the intent of this section. Are you waiting for input from the mentioned teams before finalizing this decision?

spolti: Are these changes aligned with Dashboard team?
If the user wants to provide their own certs?

`Can ServiceMesh team make the solution more efficient? `

what do you mean by this last question?

lazarotti: Indeed, @Jooho this section is not needed.
There is already the ticket for private/pubic route that will be addressed by Dashboard team (2.11, or 2.12)
"Does Serverless/ServiceMesh team support the proposed way?" is defined by the acceptance of this PR review, so it does not need to be on the ADR, it is just the review to be accepted.
"Can ServiceMesh team make the solution more efficient? " - Rob mentioned by internal docs that there is nothing to add, so his acceptance here means his is fine with the proposal. The same from Knative team.
So these questions were already answered

lazarotti (approved):

israel-hdez (approved):

rcernich (approved):

spolti (approved):

Jooho:
@zdtsw Could you please review this ADR?

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
Jooho: removed open question part.

Jooho:
@ReToCode I know you commented "approved" but could you please approve PR like other people did? (refer the last part of this PR)

zdtsw:
Approved

Based on this PR https://github.com/opendatahub-io/opendatahub-operator/pull/1056 ended to bring the templates in Operator git repo. 
As for the solution itself, I do not have much to say about.

Jooho:
@LaVLaS This ADR got all the approval from each team. Could you please review this and merge it?

rhuss (approved):
Looks good to me, thanks!

model-serving/ODH-ADR-MS-0001-kserve-private-network-in-cluster.md
```diff
@@ -0,0 +1,77 @@
...
+| Reviewed by           | Date         | Approval | Notes |
+| --------------------- | -----------  | -------- | ----- |
+| Jooho Lee             | Jun 3, 2024  | Approved |       |
+| Edgar Hernandez       | Jun 3, 2024  | Approved |       |
+| Alessandro Lazarotti  | Jun 3, 2024  | Approved |       |
+| Daniele Zonca         | Jun 3, 2024  | Approved |       |
+| Reto Lehmann          | Jun 3, 2024  | Approved |       |
+| Rob Cernich           | Jun 18, 2024 | Approved |       |
+| Lucas Fernadez aragon | Jun 12, 2024 | Approved |       |
+| Wen Zhou              | Jun 27, 2024 | Approved |       |
```
rhuss, line 77:
```suggestion
| Wen Zhou              | Jun 27, 2024 | Approved |       |
| Roland Huss           | Jun 27, 2024 | Approved |       | 
```

ReToCode (approved):
Approved

etirelli (approved):

PR #70, opened 2024-10-22
add ADR for auth CRD

StevenTobin:
## Description

Add an ADR recording the decision to add an auth CRD.
## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.

### DIFF
--- a/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
+++ b/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
@@ -40,7 +40,7 @@ Centralising auth concerns will simplify the architecture of auth features, secu
 
 ## How
 
-A new  API is added to the operator in the form of an auth CRD in the services api group. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
+A new  API is added to the operator in the form of an auth CRD in the services api group. This will initially handle a list of adminGroups and allowedGroups (migrated from the groupsConfig field of the dashboardConfig) which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
 
 The Auth CR is a singleton like the DSC and DSCi CRs.
 
@@ -52,6 +52,7 @@ metadata:
     name: odhAuth
 spec:
     adminGroups: []
+    allowedGroups: []
 ```
 
 ## Open Questions
--- a/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
+++ b/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
@@ -40,7 +40,9 @@ Centralising auth concerns will simplify the architecture of auth features, secu
 
 ## How
 
-A new API is added to the operator in the form of an auth CRD in the services api group. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
+A new  API is added to the operator in the form of an auth CRD in the services api group. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
+
+The Auth CR is a singleton like the DSC and DSCi CRs.
 
 An example of the CR:
 ```
--- a/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
+++ b/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
@@ -53,7 +53,8 @@ spec:
 ```
 
 ## Open Questions
-    
+* Migration Path:
+    *   We will use [CEL](https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutability-upon-object-creation) to make the groupsConfig field in the current OdhDashboardConfig CRD immutable. The operator will manage copying the content of the field over to the new CRD and the Auth CRD will be the new source of truth for adminGroups and allowedGroups. 
 
 ## Alternatives
 
--- a/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
+++ b/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
@@ -21,7 +21,15 @@ This document outlines the decision to implement an auth CRD in ODH.
 
 ## Why
 
-There is a growing need for the ODH operator to handle authentication for the platform. OIDC is not handled gracefully over the entire system, and there are use cases where individual components are handling auth individually which those components would like to move away from. Centralising auth concerns will simplify the architecture of auth features, security and ease of use of these features for users.
+There is a growing need for the ODH operator to provide centralized authentication and authorization services for the platform.  The near term needs are :
+
+-  Add support for managing user groups. Currently this  is  managed by the Dashboard component and needs to be centralized.
+-  Prepare for upcoming for changes to handle OCP Platform support for external OIDC authentication
+
+Longer term,  this API and its controller will be enhanced to handle additional RBAC configuration currently handled in the Dashboard component, such as creating roles and role bindings.
+
+Centralising auth concerns will simplify the architecture of auth features, security and ease of use of these features for users.
+
 
 ## Goals
 
--- a/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
+++ b/architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
@@ -32,13 +32,19 @@ There is a growing need for the ODH operator to handle authentication for the pl
 
 ## How
 
-A new API is added to the operator in the form of an *DSCAuth* (Naming to be decided) CRD. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
+A new API is added to the operator in the form of an auth CRD in the services api group. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
+
+An example of the CR:
+```
+apiVersion: services.opendatahub.io/v1alpha1
+kind: Auth
+metadata:
+    name: odhAuth
+spec:
+    adminGroups: []
+```
 
 ## Open Questions
-
-* Naming of the new CRD.
-    * DSCAuth
-    * odhAuth
     
 
 ## Alternatives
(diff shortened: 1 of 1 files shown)


lburgazzoli:
In term of naming, as part of the [RHOAIENG-10498](https://issues.redhat.com/browse/RHOAIENG-10498), we are going to introduce a new API group  `services.opendatahub.io` for shared services / concerns [RHOAIENG-13009](https://issues.redhat.com/browse/RHOAIENG-13009) so it could be something like:

```yaml
apiVersion: services.opendatahub.io/v1alpha1
name: Auth
```

zdtsw:
the ADR is for "we need a new CRD in Operator to handle auth releated work" but nothing defined what need/should be captured in this CRD for now, except it will handle adminGroup
is my understanding correct?

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,69 @@
...
+## Non-Goals
+
+
+## How
+
+A new API is added to the operator in the form of an *DSCAuth* (Naming to be decided) CRD. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
+
+## Open Questions
+
+* Naming of the new CRD.
```
etirelli, line 39: In order to move forward with this ADR, it would be best to have a decision on the name of the CRD, and a definition of the structure. Also, can you add some information on how the components would then leverage the CRD to overcome the current deficiencies?

etirelli (changes requested):
As per my comment.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,69 @@
...
+## Goals
+
+* Define a new auth CRD.
+
+## Non-Goals
+
+
+## How
+
+A new API is added to the operator in the form of an *DSCAuth* (Naming to be decided) CRD. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
```
VaishnaviHire, line 35: Will this be something that will only be included when Dashboard component is enabled?

If not, should this instead be an internal api with configuration fields exposed in DSCI?

StevenTobin:
> In term of naming, as part of the [RHOAIENG-10498](https://issues.redhat.com/browse/RHOAIENG-10498), we are going to introduce a new API group `services.opendatahub.io` for shared services / concerns [RHOAIENG-13009](https://issues.redhat.com/browse/RHOAIENG-13009) so it could be something like:
> 
> ```yaml
> apiVersion: services.opendatahub.io/v1alpha1
> name: Auth
> ```

This makes sense, I'll add that. Moving it under this services API will reduce friction later.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,75 @@
...
+| Tickets        | [RHAOIENG-14807](https://issues.redhat.com/browse/RHOAIENG-14807)|
+| Other docs:    | none |
+
+## What
+
+This document outlines the decision to implement an auth CRD in ODH.
+
+## Why
+
+There is a growing need for the ODH operator to handle authentication for the platform. OIDC is not handled gracefully over the entire system, and there are use cases where individual components are handling auth individually which those components would like to move away from. Centralising auth concerns will simplify the architecture of auth features, security and ease of use of these features for users.
```
lphiri, line 24:
```suggestion
There is a growing need for the ODH operator to provide centralized authentication and authorization services for the platform.  The near term needs are :

-  Add support for managing user groups. Currently this  is  managed by the Dashboard component and needs to be centralized.
-  Prepare for upcoming for changes to handle OCP Platform support for external OIDC authentication

Longer term,  this API and its controller will be enhanced to handle additional RBAC configuration currently handled in the Dashboard component, such as creating roles and role bindings.

Centralising auth concerns will simplify the architecture of auth features, security and ease of use of these features for users.

```

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,75 @@
...
+```
+apiVersion: services.opendatahub.io/v1alpha1
+kind: Auth
+metadata:
+    name: odhAuth
+spec:
+    adminGroups: []
+```
+
+## Open Questions
```
lphiri, line 47: Can we add discussions here about how we migrate/upgrade current dashboard implementation?
Do we deprecate  Dashboard userGroups over some period, or do we just switch right away given the Dashboard config API is alpha? 
@andrewballantyne

andrewballantyne: @csams and I were talking about that. I would recommend a cold turkey switchover; it's alpha on our side, so I don't imagine we have a reliance here in our support to migrate slowly... but I don't know how we can cleanly do that tbh. 

The cold turkey approach is done like so (I would guess):
* Upgrade available
* User invokes / auto upgrade invokes
* Operator would need to create new location (likely a new CRD & CR? or maybe just a CRD field in DSCI or something?)
* Copy the data from OdhDashboardConfig over
* Invoke Component upgrade (Dashboard at least)
* Dashboard CRD drops the field -- the CR loses the values
* We are upgraded

If the order is not pretty much what is above (given infrastructures of the operator today) then I imagine we need to slowly migrate over release(s) and I imagine many (EUS/Stable/Fast all need to have 1 with the migration effort).

cc @etirelli FYI

lphiri (commented):
Suggested some wording tweaks.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
StevenTobin: I added some text around the upgrade strategy we discussed in a seperate meeting.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,84 @@
...
+
+An example of the CR:
+```
+apiVersion: services.opendatahub.io/v1alpha1
+kind: Auth
+metadata:
+    name: odhAuth
+spec:
+    adminGroups: []
+```
```
andrewballantyne, line 53: You don't speak to it here -- is this a singleton like DSC and DSCI or would they create multiple? If they can, is there any benefit today for that or is that for "future use-cases"?

StevenTobin: Yep, this is intended to be a singleton, I've added that to the ADR now.

lphiri (approved):
LGTM

lburgazzoli:
LGTM

lburgazzoli (approved):

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,86 @@
...
+The Auth CR is a singleton like the DSC and DSCi CRs.
+
+An example of the CR:
+```
+apiVersion: services.opendatahub.io/v1alpha1
+kind: Auth
+metadata:
+    name: odhAuth
+spec:
+    adminGroups: []
```
csams, line 54: Let's include the allowedGroups from dashboard config in the example just so the intent is explicit if we're going to disable it in their CR.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,86 @@
...
+## How
+
+A new  API is added to the operator in the form of an auth CRD in the services api group. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
+
+The Auth CR is a singleton like the DSC and DSCi CRs.
+
+An example of the CR:
+```
+apiVersion: services.opendatahub.io/v1alpha1
+kind: Auth
```
csams, line 50: @lphiri @etirelli Should we combine authn and authz concerns in this CR, or should we separate them and give this one a more specific name like `Authz` or `Authorization`?

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,86 @@
...
+## Goals
+
+* Define a new auth CRD.
+
+## Non-Goals
+
+
+## How
+
+A new  API is added to the operator in the form of an auth CRD in the services api group. This will initially handle a list of adminGroups which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
```
csams, line 43: A comment below mentions disabling the ability to edit the dash config stanza that I think includes both admin group and allowed groups.  Are we doing both?

csams (changes requested):
I have a couple questions about naming and making sure the scope is clear.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
StevenTobin: :+1:  good point. I've added that.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
StevenTobin: Good point. I've made it more explicit.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
csams: If we keep the 1 CR, maybe namespace the authn/z concerns beneath the spec: `spec.authn`, `spec.authz`.

andrewballantyne (approved):

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
StevenTobin: A single CR for both authz and authn would make more sense I think than seperating them, and namespacing like you point out makes it clearer which setting in the CR relates to which concern.

I would agree for futureproofing.

architecture-decision-records/operator/ODH-ADR-Operator-0007-auth-crd.md
```diff
@@ -0,0 +1,87 @@
...
+
+## How
+
+A new  API is added to the operator in the form of an auth CRD in the services api group. This will initially handle a list of adminGroups and allowedGroups (migrated from the groupsConfig field of the dashboardConfig) which the operator will reconcile for access to dashboard UIs and applying requisite openshift permissions. This is intended to be the canonical place for auth configuration for future auth initiatives.
+
+The Auth CR is a singleton like the DSC and DSCi CRs.
+
+An example of the CR:
+```
+apiVersion: services.opendatahub.io/v1alpha1
```
zdtsw, line 49: after the lastest change this should be updated to 
`apiVersion: services.platform.opendatahub.io/v1alpha1`

lphiri (approved):

PR #73, opened 2024-11-04
Dashboard: describe annotation opendatahub.io/connection-type-ref

christianvogt:
Describe the annotation `opendatahub.io/connection-type-ref` as used to reference a connection type `ConfigMap` for a connection.

Previously we were using `opendatahub.io/connection-type` however this annotation was later found to have special meaning as it is the type transformed into the `storage-config` to be used by kserve. Therefore it must continue to hold only meaningful values for that scenario.

### DIFF
--- a/documentation/components/dashboard/k8sLabelsAndAnnotations.md
+++ b/documentation/components/dashboard/k8sLabelsAndAnnotations.md
@@ -125,11 +125,12 @@ For the Project Sharing feature specifically:
 ### Connections
 
 * Labels
-  * `opendatahub.io/managed` - helps us define that this is a managed resource by us and not a normal `Secret`
+  * `opendatahub.io/managed` - Legacy value. Identifies data connections which are watched by the model mesh controller for the purpose of populating the model serving `storage-config`
 * Annotations
   * [`openshift.io/display-name`]
   * [`openshift.io/description`]
-  * `opendatahub.io/connection-type` - a reference to the connection type; eg. `s3`
+  * `opendatahub.io/connection-type` - Legacy value. Used to identify S3-compatible data connections; `s3` is the only supported value
+  * `opendatahub.io/connection-type-ref` - a reference to the connection type that is used to create the connection
 
 ### ImageStreams
 


documentation/components/dashboard/k8sLabelsAndAnnotations.md
```diff
@@ -129,7 +129,8 @@ For the Project Sharing feature specifically:
 * Annotations
   * [`openshift.io/display-name`]
   * [`openshift.io/description`]
-  * `opendatahub.io/connection-type` - a reference to the connection type; eg. `s3`
+  * `opendatahub.io/connection-type` - the type of connection; eg. `s3`
```
andrewballantyne, line 132: If you're not S3... what are you? Now that we support multiple types and heavily dynamic ones?

documentation/components/dashboard/k8sLabelsAndAnnotations.md
```diff
@@ -129,7 +129,8 @@ For the Project Sharing feature specifically:
...
+  * `opendatahub.io/connection-type-ref` - a reference to the connection type `ConfigMap` that is used to create the connection
```
andrewballantyne, line 133: I'm not sure we need to tie this to say `ConfigMap` as `Connection Type` should be enough.

andrewballantyne (changes requested):

documentation/components/dashboard/k8sLabelsAndAnnotations.md
christianvogt: this is consumed by the modelmesh controller. I wasn't sure if we wanted to be that specific. At the moment both `opendatahub.io/connection-type` and label `opendatahub.io/managed` are specific to the usecase of modelmesh.

documentation/components/dashboard/k8sLabelsAndAnnotations.md
christianvogt: fair enough

documentation/components/dashboard/k8sLabelsAndAnnotations.md
andrewballantyne: Does this value ever change from s3? If it's only for S3 -- I think we need to update this and state that. The goals originally were that this would expand to other types (which seems not to be the case anymore 😞 )

andrewballantyne:
```suggestion
  * `opendatahub.io/connection-type` - Legacy value. Used to identify S3-compatible data connections `s3`
```
Maybe something like this?

andrewballantyne (approved):

PR #98, opened 2025-06-25
Add Gateway Inference Extension CRDs owners ADR

pierDipi:
## Description

This Architecture Decision Record (ADR) defines the ownership and
installation responsibility for the Gateway Inference Extension (GIE)
CRDs within Open Data Hub (ODH).

Rendered view https://github.com/pierdipi/architecture-decision-records/blob/gateway-inference-extension-crds-owners/architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md

## Summary by CodeRabbit

- **Documentation**
  - Added an Architecture Decision Record clarifying who owns and manages Gateway Inference Extension (GIE) components and CRDs in Open Data Hub.
  - Documents how shared dependencies (like InferencePool) are detected and reconciled across teams, and how components should behave if those CRDs are absent or removed.
  - Notes risks, stakeholder impacts, and coordination guidance to support independent evolution of components.


architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
```diff
@@ -0,0 +1,97 @@
...
+This Architecture Decision Record (ADR) defines the ownership and installation responsibility for the Gateway Inference
+Extension (GIE) CRDs within Open Data Hub (ODH).
+
+## Why
+
+The Gateway Inference Extension (GIE) introduces several CRDs. While most are leveraged by the GIE scheduler (
+llm-d-inference-scheduler), which is owned by OpenDataHub, the InferencePool CRD presents a cross-team dependency. It is
+used by both the llm-d-inference-scheduler and the OpenShift Container Platform (OCP) Networking team's Gateway API
+implementation (Istio) to configure the Envoy ExtProc. This shared dependency necessitates a clear ownership model to
+avoid installation conflicts and ensure a clear lifecycle management.
```
bartoszmajsak, line 25:
```suggestion
avoid installation conflicts and ensure a predicable lifecycle management.
```

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
```diff
@@ -0,0 +1,97 @@
...
+  their components can utilize the newer CRD version in future updates.
+- If the OCP Networking team's Istio implementation requires a different version of the InferencePool CRD than what
+  OpenDataHub installs, it could lead to feature incompatibility or unexpected behavior. This risk is mitigated by clear
+  communication channels between the teams and robust integration testing.
+- Users of the OCP Gateway API who wish to use InferencePool will have a dependency on OpenDataHub being installed.
+  This is considered an acceptable trade-off, as the primary use case for GIE is driven by the model serving
+  capabilities provided by ODH.
+- CRD Removal: If OpenDataHub is uninstalled, the GIE CRDs will be removed. This could negatively impact the OCP
+  Networking components if they are still configured to use them. The OCP Gateway controller must be robust enough to
+  handle the absence of the CRD gracefully.
```
bartoszmajsak, line 83: I think Opendatahub Operator does not remove CRDs when uninstalled. That actually matches my expectations the cluster/components should be left in the functional state after removing the operator.

If we really want to do that though, then it might be tricky https://github.com/kubernetes/kubernetes/issues/79610

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
```diff
@@ -0,0 +1,97 @@
...
+
+- If the InferencePool CRD evolves with breaking changes, a coordinated update process is required. OpenDataHub must
+  ensure the installed CRD version is compatible with the OCP version in use, while the OCP Networking team must ensure
+  their components can utilize the newer CRD version in future updates.
+- If the OCP Networking team's Istio implementation requires a different version of the InferencePool CRD than what
+  OpenDataHub installs, it could lead to feature incompatibility or unexpected behavior. This risk is mitigated by clear
+  communication channels between the teams and robust integration testing.
+- Users of the OCP Gateway API who wish to use InferencePool will have a dependency on OpenDataHub being installed.
+  This is considered an acceptable trade-off, as the primary use case for GIE is driven by the model serving
+  capabilities provided by ODH.
```
bartoszmajsak, line 80: Isn't enforcing to install RHOAI to use GIE CRD on OpenShift too hard requirement? I would assume some customers would like to use other providers on their cluster. It's not clear to me what we really want to say here.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
```diff
@@ -0,0 +1,97 @@
...
+- Timeline for TP at risk
+
+## Risks
+
+- If the InferencePool CRD evolves with breaking changes, a coordinated update process is required. OpenDataHub must
+  ensure the installed CRD version is compatible with the OCP version in use, while the OCP Networking team must ensure
+  their components can utilize the newer CRD version in future updates.
+- If the OCP Networking team's Istio implementation requires a different version of the InferencePool CRD than what
+  OpenDataHub installs, it could lead to feature incompatibility or unexpected behavior. This risk is mitigated by clear
+  communication channels between the teams and robust integration testing.
```
bartoszmajsak, line 77: > robust integration testing

There's a TCK that is being worked on by OSSM team, I think that can be a solid first step.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
```diff
@@ -0,0 +1,97 @@
...
+must be robust enough to handle the absence of the CRD gracefully.
+
+![proposed ownership and flow](./images/ODH-ADR-MS-0002-gateway-inference-extension-resource-diagram.jpg)
+
+## Alternatives
+
+### OCP Networking Owns the CRDs
+
+The OCP Networking team, as owners of the Gateway API implementation, could own the installation of the `InferencePool`
+CRD, while leaving the remaining CRDs to OpenDataHub.
```
bartoszmajsak, line 57: > while leaving the remaining CRDs to OpenDataHub.

Would it make sense to list those CRDs for clarity?

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
shaneutt, line 80: Currently, we have no mechanism to block users from deploying and managing these CRDs themselves. You'll **want to have a process for how to absorb the responsibility for any existing CRD**. You'll also **want to protect the CRD from tampering once you deploy it**. You can use a **Validating Admission Policy (VAP)** for this.

However, given how narrow the use cases are for GIE I would anticipate this being a problem would be _very rare_. So it may be reasonable to have your process be something like this:

- In your controller startup logic, detect whether the CRD exists
  - if NO, then you're golden. deploy your CRD and a Validating Admission Policy
  - if YES, determine if it was deployed by you (check VAP, annotations, etc)
    - if YES do any upgrades you need, etc and you're all set
    - if NO throw an alert and push to a support interaction

Where the conflict situation is not automated, but rather just pushes for a support interaction to resolve manually (and then it would be very interesting to see if it actually ever comes up).

This process has similarities with how OCP 4.18 -> 4.19 upgrades check for the core Gateway API CRDs, and how we apply a VAP to protect them from tampering. See [here](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/gateway-api-crd-life-cycle-management.md#crd-management-succession) for the prior art.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
```diff
@@ -0,0 +1,97 @@
...
+- If the InferencePool CRD evolves with breaking changes, a coordinated update process is required. OpenDataHub must
+  ensure the installed CRD version is compatible with the OCP version in use, while the OCP Networking team must ensure
+  their components can utilize the newer CRD version in future updates.
+- If the OCP Networking team's Istio implementation requires a different version of the InferencePool CRD than what
+  OpenDataHub installs, it could lead to feature incompatibility or unexpected behavior. This risk is mitigated by clear
+  communication channels between the teams and robust integration testing.
+- Users of the OCP Gateway API who wish to use InferencePool will have a dependency on OpenDataHub being installed.
+  This is considered an acceptable trade-off, as the primary use case for GIE is driven by the model serving
+  capabilities provided by ODH.
+- CRD Removal: If OpenDataHub is uninstalled, the GIE CRDs will be removed. This could negatively impact the OCP
```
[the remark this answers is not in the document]
shaneutt: Right. Removing the CRD has the implication of causing downtime for live inference traffic for any other implementation that may be in play on the cluster. Given those implications, you may want to put the responsibility on a cluster admin to do this manually, documenting the need to understand the implication and assess any risks to live systems on the cluster. I strongly advise against automated mechanisms to remove a CRD.

shaneutt (commented):
@JoelSpeed you might find this interesting (CRD fun)

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
```diff
@@ -0,0 +1,97 @@
...
+used by both the llm-d-inference-scheduler and the OpenShift Container Platform (OCP) Networking team's Gateway API
+implementation (Istio) to configure the Envoy ExtProc. This shared dependency necessitates a clear ownership model to
+avoid installation conflicts and ensure a clear lifecycle management.
+
+## Goals
+
+- To establish a single, authoritative owner for the installation and management of the Gateway Inference Extension
+  CRDs.
+- To ensure a seamless user experience where features dependent on these CRDs are enabled dynamically.
+
```
shaneutt, line 32: Not sure where the exact right place to add it is, but _somewhere_ in this document you may want to point out the future consideration of moving responsibility for these CRDs to the platform as they mature. Having control of them now makes a lot of sense because it allows versatility and speed as new additive changes come in during the first year or so of GA, but once things settle down and mature over the next couple years there are benefits to having it managed by the platform.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
pierDipi: we've added a link as part of https://github.com/opendatahub-io/architecture-decision-records/pull/98#discussion_r2166711581

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
pierDipi: I've added a "Future considerations" section as part of the alternative "OCP Networking Owns the CRDs"

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
bartoszmajsak: Link is enough.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
```diff
@@ -0,0 +1,97 @@
...
+## How
+
+OpenDataHub, through its Model Serving components, will be responsible for installing and managing the lifecycle of all
+Gateway Inference Extension (GIE) CRDs, including InferencePool.
+
+The OCP Gateway API implementation (Istio) will be designed to detect the presence of the InferencePool CRD on the
+cluster. Its controller will dynamically enable the logic to reconcile InferencePool resources only when the CRD is
+available. This "watch-based" enablement decouples the OCP Networking components from the installation of the GIE CRDs,
+allowing each component to evolve independently. If OpenDataHub is uninstalled, the GIE CRDs will be removed. This could
```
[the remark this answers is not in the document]
pierDipi: maybe even if the CRDs are left as zombies, don't we still want to have Istio handle such cases? admins can still remove the CRDs if they want/need to and I rather have Istio continue working and recover if those CRDs are re-installed at later point that fail completely

Perhaps I can reword the sentence to stay generic on what removes CRDs?

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
pierDipi: I expect that this is potentially a problem with any CRD we install.

cc @danielezonca

architecture-decision-records/model-serving/ODH-ADR-MS-0002-gateway-inference-extension-crds-owners.md
pierDipi: Even if it's not done by Operator manually, don't we still want the Gateway API controller (Istio) to handle that case gracefully?

pierDipi:
@shaneutt @danielezonca @bartoszmajsak the one open point I see here is: if we want to have and agree on having the GW API implementation (istio) gracefully handle gracefully the case where the CRD is removed (by whatever actor: admin, ODH operator, etc)

refs:
- https://github.com/opendatahub-io/architecture-decision-records/pull/98#discussion_r2166717352
- https://github.com/opendatahub-io/architecture-decision-records/pull/98#discussion_r2166694217
- https://github.com/opendatahub-io/architecture-decision-records/pull/98#discussion_r2166576305

danielezonca:
> @shaneutt @danielezonca @bartoszmajsak the one open point I see here is: if we want to have and agree on having the GW API implementation (istio) gracefully handle the case where the CRD is removed (by whatever actor: admin, ODH operator, etc)
> 
> refs:
> 
> * [Add Gateway Inference Extension CRDs owners ADR #98 (comment)](https://github.com/opendatahub-io/architecture-decision-records/pull/98#discussion_r2166717352)
> * [Add Gateway Inference Extension CRDs owners ADR #98 (comment)](https://github.com/opendatahub-io/architecture-decision-records/pull/98#discussion_r2166694217)
> * [Add Gateway Inference Extension CRDs owners ADR #98 (comment)](https://github.com/opendatahub-io/architecture-decision-records/pull/98#discussion_r2166576305)

I would probably move istio gracefully management of the CRD removal in a future design decision: overall we should apply the same policy we do for all other CRDs in ODH/RHOAI that are not removed when we remove the operator/controller.

dgn:
What's the reason for having to handle deletions? CRD updates?

danielezonca (approved):
/lgtm

PR #131, opened 2026-04-21
Add MaaS tenant configuration custom resource ADR

lphiri:
## Description
Add ADR to record MaaS code migration from Operator to component repository.

## Merge criteria:

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record defining a dual-CR model-serving approach: a user-facing, namespace-scoped Tenant resource and an internal ModelsAsService resource for operator integration.
  * Describes automatic bootstrapping of a default-tenant singleton, CEL-based enforcement, and a label-based strategy for tracking tenant-owned resources across namespaces.
  * Covers goals, non-goals, alternatives considered, security/RBAC implications, transition risks, and stakeholder impacts.

### DIFF
--- a/architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
+++ b/architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
@@ -1,0 +1,131 @@
+# ODH-ADR-MS-0002: MaaS Tenant CR Introduction for User Configuration
+
+|                |            |
+| -------------- | ---------- |
+| Date           | 2026-04-21 |
+| Scope          | Model Serving |
+| Status         | Draft |
+| Authors        | Lindani Phiri |
+| Supersedes     | N/A |
+| Superseded by: | N/A |
+| Tickets        | |
+| Other docs:    | [ODH Operator PR #3412](https://github.com/opendatahub-io/opendatahub-operator/pull/3412), [MaaS PR #735](https://github.com/opendatahub-io/models-as-a-service/pull/735) |
+
+## What
+
+Introduction of a new `MaaSTenant` custom resource for user-facing Models-as-a-Service platform configuration while retaining the existing `ModelsAsService` CR for internal operator integration.
+
+## Why
+
+This architectural change addresses several design and operational concerns:
+
+- **ODH Operator Design Conformance**: Align with ODH operator design principles by separating internal operator integration (`components.platform.opendatahub.io`) from user-facing configuration
+- **Clear API Separation**: Establish distinct APIs for operator integration vs. user configuration, improving clarity and reducing coupling
+- **Future Migration Preparation**: Prepare for eventual transition from cluster-scoped to namespace-scoped configuration with appropriate RBAC constraints  
+- **Component Modularity**: Locate component-specific user configuration within the component itself, reducing coupling and improving modularity
+- **Strategic Timing**: Current release has no existing user-facing configuration to migrate, allowing introduction of new user-facing configuration without future migration burden
+
+## Goals
+
+* Introduce namespace-scoped `MaaSTenant` CR for user-facing configuration
+* Retain `ModelsAsService` CR for internal ODH operator integration
+* Enable self-bootstrapping MaaS platform via `maas-controller`
+* Establish clear separation between operator integration and user configuration APIs
+* Implement singleton enforcement to maintain current behavior
+* Support namespace-scoped multi-tenancy architecture for future enhancement
+* Maintain full backward compatibility with existing operator integration
+
+## Non-Goals
+
+* Complete multi-tenancy architecture implementation (requires RHOAI-level multi-tenant architecture completion first)
+* Immediate production multi-tenant support (application remains single-tenant but uses tenant-ready naming conventions)
+* Breaking changes to existing MaaS functionality during migration
+
+## How
+
+The implementation involves a dual-CR architecture across the ODH Operator and MaaS repositories:
+
+### ODH Operator Changes
+- Retain existing `ModelsAsService` controller and CRD for internal operator integration
+- Update component handling to support both internal (`ModelsAsService`) and external (`Tenant`) CRs
+- Maintain `DataScienceCluster` integration through existing `ModelsAsService` CR
+- Add operator install manifests for the new MaaS component integration
+
+### Implementation Approach
+
+The migration moves implementation from the common platform API to the dedicated maas.opendatahub.io API:
+
+* **Tenant CR Creation**: Establish a namespace-scoped `Tenant` Custom Resource that includes:
+  - API key configuration  
+  - External OIDC settings
+  - Gateway configuration
+  - Telemetry policies
+* **Singleton Pattern**: Enforce single tenant via CEL validation requiring name `default-tenant`
+* **Self-Bootstrapping**: maas-controller creates the default Tenant automatically
+* **Cross-Namespace Resource Management**: Use tracking labels (`maas.opendatahub.io/tenant-name` and `maas.opendatahub.io/tenant-namespace`) for resources across different namespaces
+
+
+### Dual-CR Architecture
+- `ModelsAsService` CR (`components.platform.opendatahub.io`): Internal operator integration, cluster-scoped
+- `Tenant` CR (`maas.opendatahub.io`): User-facing configuration, namespace-scoped
+- MaaS controller reconciles user configuration from `Tenant` CR independently
+- ODH operator continues to manage component lifecycle through `ModelsAsService` CR
+
+## Alternatives
+
+### Alternative 1: Single CR Approach (ModelsAsService only)
+**Pros**: Simpler architecture, single API surface, no additional complexity
+**Cons**: Continues using internal ODH operator CRs for user configuration, doesn't prepare for future multi-tenancy, maintains tight coupling between operator and component concerns, violates separation of internal vs user-facing APIs
+
+### Alternative 2: Complete ModelsAsService Replacement
+**Pros**: Clean break, single user-facing API, removes internal CR complexity
+**Cons**: Breaks existing operator integration patterns, requires migration of all ODH operator component handling, higher risk and complexity for operator changes
+
+### Alternative 3: Cluster-scoped Tenant CR  
+**Pros**: Simpler RBAC, consistent with current ModelsAsService scope, easier cross-namespace access
+**Cons**: CRD scope is immutable, would require complex migration path for future namespace-scoped multi-tenancy, doesn't align with future architecture direction, limits RBAC isolation
+
+### Alternative 4: Enhanced ModelsAsService CR
+**Pros**: Leverages existing CR, no new API complexity, maintains operator integration
+**Cons**: Mixes internal operator concerns with user configuration, harder to evolve independently, doesn't prepare for namespace-scoped multi-tenancy
+
+## Security and Privacy Considerations
+
+- Namespace-scoped CRs enable future RBAC-constrained tenant isolation
+- Cross-namespace resource management requires careful label-based tracking to prevent privilege escalation
+- Self-bootstrapping mechanism must ensure appropriate security context for tenant creation
+
+## Risks
+
+* **Migration Complexity**: Risk of configuration inconsistencies during the transition period
+* **Cross-Namespace Resource Management**: Potential for resource orphaning if tracking labels are not properly managed
+* **Terminology Confusion**: Using "Tenant" terminology before full multi-tenancy implementation may confuse operators and developers
+
+
+## Stakeholder Impacts
+
+| Group                         | Key Contacts     | Date       | Impacted? |
+| ----------------------------- | ---------------- | ---------- | --------- |
+| Platform Core Operator Team            | TBD              | TBD        | Yes |
+| MaaS/Model Serving Team      | TBD              | TBD        | Yes |
+| Documentation Team           | TBD              | TBD        | Yes |
+| QE/Testing Team             | TBD              | TBD        | Yes |
+
+**Impact Details:**
+- **ODH Operator Team**: Minor changes to support dual-CR architecture, existing ModelsAsService integration preserved
+- **MaaS/Model Serving Team**: New controller implementation for Tenant CR, additional reconciliation architecture  
+- **Documentation Team**: New user configuration documentation, installation guide updates
+- **QE/Testing Team**: Additional test scenarios for Tenant CR lifecycle, existing operator integration tests preserved
+
+## References
+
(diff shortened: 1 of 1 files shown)


lphiri:
@ishitasequeira  @mariusdanciu  @jland-redhat @danielezonca  @lburgazzoli

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+
+- **ODH Operator Design Conformance**: Align with ODH operator design principles by separating internal operator integration (`components.platform.opendatahub.io`) from user-facing configuration
+- **Clear API Separation**: Establish distinct APIs for operator integration vs. user configuration, improving clarity and reducing coupling
+- **Future Migration Preparation**: Prepare for eventual transition from cluster-scoped to namespace-scoped configuration with appropriate RBAC constraints  
+- **Component Modularity**: Locate component-specific user configuration within the component itself, reducing coupling and improving modularity
+- **Strategic Timing**: Current release has no existing user-facing configuration to migrate, allowing introduction of new user-facing configuration without future migration burden
+
+## Goals
+
+* Introduce namespace-scoped `MaaSTenant` CR for user-facing configuration
```
mariusdanciu: Why end users should be able to configure maas-api behavior per namespace ?

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+* Retain `ModelsAsService` CR for internal ODH operator integration
```
mariusdanciu, line 31: Having two separate CRs worries me a bit. It seems like an incomplete or temporary design.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+* Immediate production multi-tenant support (application remains single-tenant but uses tenant-ready naming conventions)
+* Breaking changes to existing MaaS functionality during migration
+
+## How
+
+The implementation involves a dual-CR architecture across the ODH Operator and MaaS repositories:
+
+### ODH Operator Changes
+- Retain existing `ModelsAsService` controller and CRD for internal operator integration
+- Update component handling to support both internal (`ModelsAsService`) and external (`Tenant`) CRs
```
mariusdanciu: This worries me. For an admins seeing both CRs can be quite confusing.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Implementation Approach
+
+The migration moves implementation from the common platform API to the dedicated maas.opendatahub.io API:
+
+* **Tenant CR Creation**: Establish a namespace-scoped `Tenant` Custom Resource that includes:
+  - API key configuration  
+  - External OIDC settings
+  - Gateway configuration
+  - Telemetry policies
```
mariusdanciu: If the Tenant CR is meant to be specfic to a maas-api service instance (and this has nothing to do with actual tenant namespaces) meaning that it is a maas-api service instance config then I strongly suggest to rename this to perhaps something like MaaSService CR or similar. To me this can be very confusing in a multi-tenant context. But to be honest I don't see what is wrong with the current ModelAsService CR.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+* **Singleton Pattern**: Enforce single tenant via CEL validation requiring name `default-tenant`
```
mariusdanciu: Why ? If the user wants to create 5 maas-api service instance, they should be able to create 5 CR objects. However we should recommend customers to create a single maas-api instance in the control plan because a single  maas-api instance will be able to handle multiple tenants in the same time.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+* **Self-Bootstrapping**: maas-controller creates the default Tenant automatically
```
mariusdanciu, line 64: For sure there should be a default maas-api (as today) but we should not call this a tenant. I don't think that an actual tenant requires a CR. A tenant is simply the user namespace name, just like in mlflow.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+* **Cross-Namespace Resource Management**: Use tracking labels (`maas.opendatahub.io/tenant-name` and `maas.opendatahub.io/tenant-namespace`) for resources across different namespaces
+
+
+### Dual-CR Architecture
```
mariusdanciu, line 68: This worries me as mentioned above.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+- `ModelsAsService` CR (`components.platform.opendatahub.io`): Internal operator integration, cluster-scoped
+- `Tenant` CR (`maas.opendatahub.io`): User-facing configuration, namespace-scoped
```
mariusdanciu, line 70: I don't think we should call this a Tenant if it is not about tenancy.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+- MaaS controller reconciles user configuration from `Tenant` CR independently
+- ODH operator continues to manage component lifecycle through `ModelsAsService` CR
+
+## Alternatives
+
+### Alternative 1: Single CR Approach (ModelsAsService only)
```
mariusdanciu, line 76: I still prefer **Altenative 1.**  I don't see how moving away from this prepares us for multi-tenancy. Multi-tenancy should not require a new CR. 

I don't understand 
> maintains tight coupling between operator and component concerns, violates separation of internal vs user-facing APIs

Having a CR that simply holds the configuration properties for a maas-api instance keeps things very simple and intuitive. I don't see how/why this affects multi-tenancy. This is the same pattern we have for mlflow and eval-hub and there was never a need for a separate tenant CR.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+### Alternative 4: Enhanced ModelsAsService CR
+**Pros**: Leverages existing CR, no new API complexity, maintains operator integration  
+**Cons**: Mixes internal operator concerns with user configuration, harder to evolve independently, doesn't prepare for namespace-scoped multi-tenancy
+
+## Security and Privacy Considerations
+
+- Namespace-scoped CRs enable future RBAC-constrained tenant isolation
+- Cross-namespace resource management requires careful label-based tracking to prevent privilege escalation
+- Self-bootstrapping mechanism must ensure appropriate security context for tenant creation
+
```
mariusdanciu, line 97: I am still missing some context here. Say I have a maas-api service instance in the control-plane namespace. **Just one.** 

1. In my org I needs 3 tenants, therefore I create 3 separate namespaces.
2. Am I required to create a separate Tenant CR in each of these 3 namespaces and maas-api service from the control plane namespace would watch these CRs ? If so why do we need this complexity ?

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
```diff
@@ -0,0 +1,131 @@
...
+- **Clear API Separation**: Establish distinct APIs for operator integration vs. user configuration, improving clarity and reducing coupling
+- **Future Migration Preparation**: Prepare for eventual transition from cluster-scoped to namespace-scoped configuration with appropriate RBAC constraints  
+- **Component Modularity**: Locate component-specific user configuration within the component itself, reducing coupling and improving modularity
+- **Strategic Timing**: Current release has no existing user-facing configuration to migrate, allowing introduction of new user-facing configuration without future migration burden
+
+## Goals
+
+* Introduce namespace-scoped `Tenant` CR for user-facing configuration
+* Retain `ModelsAsService` CR for internal ODH operator integration
+* Enable self-bootstrapping MaaS platform via `maas-controller`
```
mariusdanciu, line 32: When I create a Tenant CR in namespace N1, will maas-controller automatically create a maas-api service instance in namespace N1 ? If so this Tenant CR is nothing more than a config object no different that ModelAsService CR..

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: @mariusdanciu  I think the naming is causing the confusio.     The way to think about this CR is that is a tenant configuration, not a config for the entire maas-api.   Closest  analogy would be  MLflowConfig

lphiri: @mariusdanciu  I think the naming is causing the confusio.     The way to think about this CR is that is a tenant configuration, not a config for the entire maas-api.   Closest  analogy would be  MLflowConfig

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: The ModelsAsService  CR is meant to configure controller behavior  (MLFLow Analogy  is MLflowOperator)  and is what a cluster admin would use to  enable or disable MaaS clusterwide ( and also set cluster wide configuration).    The other CR is what a Tenant admin would use to influence their tenant specific configuration.
You are correct that the design is temporary/incomplete.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: Hopefully the previous [comment](https://github.com/opendatahub-io/architecture-decision-records/pull/131#discussion_r3124762923) addresses this one

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: It is not meant to be specific to a single maap-api service,  and I see /agree that the naming is problematic.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: This is a temporary measure for 3.4

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: Acknowledge the naming is confusing.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: Let me reword this one a bit :)

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: As pointed out earlier, the intent was not have a maa-api instance per config CR.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: No,   my expectation is that one maas-api controller  will handle all tenant CRs.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
lphiri: This will allow a cluster admin to delegate tenant specific configuration to a tenant admin.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
mariusdanciu: I did not say otherwise. Is there a 1-1 mapping between Tenant CR and maas-api instance ? In other words if I create a Tenant CR will maas-controller create a new maas-api service deployment ?

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
mariusdanciu: Not yet.

architecture-decision-records/model-serving/ODH-ADR-MS-0002-maas-tenant-cr-introduction.md
mariusdanciu: Are you assuming that each tenant will have a separate maas-api service instance and a seprate DB ? ... I don't think we can assume this.

mariusdanciu (approved):
As timing does not allow us to rename the Tenant CR to something more intuitive like maybe TenantConfig in the maas admin namespace, we should "stress" in the documentation the intent for the Tenant CR.

lburgazzoli (approved):

PR #138, opened 2026-05-27
AI-Gateway/MaaS multi tenancy

mariusdanciu:
## Description
Propose the multi-tenancy architecture for MaaS.

## How Has This Been Tested?
This ADR has been discussed on slack with other architected and lead engineers.

## Merge criteria:

## Summary by CodeRabbit

* **Documentation**
  * ADR for multi-tenant AI Gateway: one gateway per tenant with tenant-specific hostnames and per-tenant control/data-plane resources.
  * Per-tenant API instances with dedicated routes, tenant-scoped identity, and model discoverability gated by tenant admin authorization/subscription.
  * Database isolation for API keys (tenant_id, backfill, uniqueness, tenant-scoped queries), staged migration with default tenant, provisioning/error states, observability, and GDPR-aligned key retention.


architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+- Per-tenant physical database instances (logical isolation sufficient for v1)
+
+## How
+
+### Architecture Overview
+
+The multi-tenant architecture follows a **one-gateway-per-tenant** pattern. Each tenant gets:
+
+1. An AITenant CR in the `ai-tenants` namespace (cluster-level registry)
+2. A dedicated tenant admin namespace (`ai-tenant-{tenant-name}`)
```
jrhyness: This imposes restrictions on `tenant-name`.   `tenant-name` is also in the hostname below. 

We will need to limit the format of a name. 

eg
```
// +kubebuilder:validation:MaxLength=53
// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
```

?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+3. A dedicated Gateway CR (gateway.networking.k8s.io/v1) with separate Envoy pods
+4. A dedicated maas-api service instance with its own HttpRoute
+5. A MaasConfig CR for tenant-specific configuration
```
jrhyness: There is an existing Config CR with shortname maasconfig.

```
$ oc get maasconfig -o yaml
apiVersion: v1
items:
- apiVersion: maas.opendatahub.io/v1alpha1
  kind: Config
  metadata:
    creationTimestamp: "2026-05-27T12:56:00Z"
    generation: 1
    name: default
    resourceVersion: "26643"
    uid: c9f00260-d72f-4474-80c1-c4b5f6284221
  spec: {}
kind: List
metadata:
  resourceVersion: ""
```

Could we rename MaasConfig to MaasTenantConfig?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+### Control Plane
+
+#### Tenant Lifecycle Management (RHOAI Admin Role)
+
+##### Tenant Creation
+
+**Workflow**:
+
+1. RHOAI admin creates an AIGateway CR in the `ai-tenants` namespace
+2. maas-controller reconciles the CR and performs the following:
```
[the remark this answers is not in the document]
jrhyness: Valid.  AITenant is the final name @mariusdanciu ?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+
+-- After (multi-tenant)
+SELECT * FROM api_keys WHERE key_hash = ? AND tenant_id = ?;
+```
+
+### Migration Strategy
+
+#### Pre-Upgrade State
+
+- Single `model-as-a-service` namespace with one Gateway
```
jrhyness: nit:  The namespace is `models-as-a-service`  ( plural models )

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+
+-- Make tenant_id required
+ALTER TABLE api_keys ALTER COLUMN tenant_id SET NOT NULL;
+
+-- Add unique constraint: key hash must be unique within tenant
+-- (allows same key pattern across tenants if needed, though not recommended)
+ALTER TABLE api_keys ADD CONSTRAINT uk_api_keys_hash_tenant UNIQUE (key_hash, tenant_id);
+```
+
+**subscriptions table** (if applicable):
```
jrhyness: Not applicable.  There is no subscriptions table.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+
+#### Tenant Lifecycle Management (RHOAI Admin Role)
+
+##### Tenant Creation
+
+**Workflow**:
+
+1. RHOAI admin creates an AIGateway CR in the `ai-tenants` namespace
+2. maas-controller reconciles the CR and performs the following:
+   - Creates tenant admin namespace: `ai-tenant-{tenant-name}`
```
jrhyness, line 70: Are the tenant namespaces labelled for discovery?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+
+1. Add `tenant_id` column to `api_keys` and `subscriptions` tables
+2. Create indexes
+3. Backfill `tenant_id='default'` for all existing records
+4. Apply NOT NULL constraint
+
+**Phase 2: Create Default Tenant** (automatic on first reconciliation)
+
+1. Operator detects existing `model-as-a-service` namespace
+2. Creates AIGateway CR named `default` in `ai-tenants` namespace:
```
jrhyness: I know this is the Upgrade Process section, so maybe the wrong place to ask, but on a clean install, do we create a default tenant, or start with no tenants?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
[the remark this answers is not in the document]
mariusdanciu: fixed.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: redteam-oidc-secret
+  namespace: ai-tenants
+stringData:
+  client-secret: "847528ry8ery97qrt36"
+```
```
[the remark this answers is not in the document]
mariusdanciu: this literal is not a valid secret. Is is just a random string for exemplification.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+
+2. maas-controller copies secret to tenant admin namespace during provisioning
+3. maas-api pods mount secret from tenant namespace
+4. **Secret rotation**: Update in `ai-tenants` namespace → controller detects change → updates copy → pods restart
```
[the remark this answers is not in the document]
mariusdanciu: fixed in the lates commit.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,581 @@
...
+- **Transmission**: TLS only
+- **Scope**: Limited to creating tenant
+- **Revocation**: Immediate (no caching)
+
+### Cross-Tenant Access Prevention
+
+**Database-level**:
+- All queries filter by `tenant_id`
+- Unique constraint on `(key_hash, tenant_id)` prevents key reuse across tenants
+
```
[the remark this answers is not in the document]
mariusdanciu: apikkey are long hashes generated by the service. It is highly unlikely that the same apikey to exist for multiple tenants

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: Yes the tenant name needs to be DNS compatible names.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: I wasn't aware of this conflict. Yes I will update the CR name. Thanks !

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: oh claude, oh claude.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: Yes this is the intent we discussed in slack but I will make this explicit.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: The current model-as-a-service namespace should be the default tenant in order to be backward compatible. I think this is called out in the `Create Default Tenant ` section.

jrhyness: The `Create Default Tenant` section is under the Upgrade Process.    But now I have more questions. :) 

1.  Just to be clear, on _install_ , will the operator create a default tenant?
2.  The default tenant will have two namespaces - `models-as-a-service` and `ai-tenant-default` ?  ( Both via install and upgrade, or is the split only on upgrade?)
3.  On upgrade, MaaSSubscriptions and MaaSAuthPolicies stay in `models-as-a-service`.   Will new ones need to be created in `models-as-a-service`  for a default tenant?  Or do we need to look for them in both namespaces?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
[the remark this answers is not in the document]
mariusdanciu: >RBAC specifics for maas-controller / maas-api SAs — The RBAC section (Line 493) only covers tenant-admin and user RoleBindings. It says nothing about constraining the maas-controller ServiceAccount or the maas-api pod SA to secret/get in ai-tenants namespace only (no ClusterRole with broad secrets/*).

This ADR does not re-discuss the entire MaaS architecture. authn/z is solved at Authorino level.

>Rotation sync guarantees — The clientSecretRef on the AITenant CR (around Line 89–91) still points to a secret in ai-tenants. There's no documented behavior for what happens when that secret is rotated (detection, propagation, pod restart semantics).

There is no clientSecretRef anymore.

? Per-operation audit requirements — The Audit Logging section (Line 471) covers tenant.create / admin operations only. There's no requirement to emit an audit event for each secret read or credential access.

There are no secrets involved here anymore. Not at this point at least.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: > Just to be clear, on install , will the operator create a default tenant?
The `models-as-a-service` will become the default tenant. I added more information in the doc including the upgrade section

> The default tenant will have two namespaces - models-as-a-service and ai-tenant-default ? ( Both via install and upgrade, or is the split only on upgrade?)

There is really no `ai-default` tenant.

>On upgrade, MaaSSubscriptions and MaaSAuthPolicies stay in models-as-a-service. Will new ones need to be created in models-as-a-service for a default tenant? Or do we need to look for them in both namespaces?

As per the previous point these CRs will remain unchanged. Now I explicitly called these out in the upgrade process.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,546 @@
...
+    clientId: ai-tenant-redteam
+  
+  # Domain for this tenant (used for hostname-based routing)
+  domain: redteam.ai-gateway.apps.example.com
+  
+  # Optional TLS configuration
+  tls:
+    certificateRef:
+      name: redteam-tls-cert
+      namespace: ai-tenants
```
danielezonca, line 101: A user can configure at DSCI level a common certificate to use but I don't remember if we use it also for external endpoints.

@lphiri @lburgazzoli 
Do you remember how the injection of the certificate works?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,546 @@
...
+##### Tenant Deletion
+
+**Workflow**:
+
+1. Admin deletes the AITenant CR from the `ai-tenants` namespace
+2. maas-controller reconciles deletion and performs the following:
+   - Deletes the Gateway CR (triggers Envoy pod deletion)
+   - Deletes the tenant admin namespace (cascades to MaasTenantConfig, HttpRoute, maas-api deployment)
+   - Updates database records: marks tenant as deleted (soft delete)
+   - User namespace CRs (LlmInferenceService, HttpRoute) are **not** deleted
```
danielezonca: So what does it happen to the model? It remains deployed but not exposed?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,546 @@
...
+   - KServe controller updates HttpRoute status to `NotReady` or `Conflicted`
```
danielezonca: @pierDipi 
Is this something that is then reported in the UI already? Or we need to raise this as a gap?
I assume this is enough to make the LLMInferenceService status failing and to expose it to UI

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,546 @@
...
+##### Option 1: Shared maas-api with Multi-Gateway HttpRoute
+
+- **Approach**: Single maas-api service with one HttpRoute attached to all tenant Gateways
+- **PROS**: Lower cluster footprint, single deployment
+- **CONS**: 
+  - Complex reconciliation (update HttpRoute on every tenant create/delete)
+  - Must extract tenant from hostname at runtime (proxy header or service parsing)
+  - Complex authentication (each tenant may have different OIDC provider)
+
+##### Option 2: Dedicated maas-api per Tenant (RECOMMENDED)
```
danielezonca: Make the decided option first

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,546 @@
...
+
+#### Rollback Procedure
+
+If upgrade fails during Phase 1 or 2:
+
+1. Drop `tenant_id` column from database (if backfill incomplete)
+2. Delete `ai-tenants` namespace (if created)
+3. Restore original Gateway configuration
+
+**Note**: Rollback not supported after users create additional tenants (data loss risk).
```
danielezonca, line 405: @lburgazzoli 
Does the new modularization effort of the operator supports rollback?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,546 @@
...
+
+**Kubernetes RBAC**:
+- Tenant admins have RoleBindings only in their tenant namespace
+- Users cannot list/read other tenants' MaasAuthPolicy or MaasSubscription CRs
+
+**Gateway-level**:
+- HttpRoutes attached only to tenant Gateway
+- No cross-tenant routing possible (separate Envoy pods)
+
+### Data Deletion (GDPR Compliance)
```
danielezonca: In general we don't provide specific compliance certification so I'm fine with this proposal but I would remove the mentioning of GDPR, we cannot claim this to be enough to be complaint with GDPR

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,546 @@
...
+| Dashboard team | TBD | TBD | Yes | UI must support tenant selection and switching |
+| Operator team | TBD | TBD | Yes | maas-controller must reconcile AITenant CRs |
+| KServe/Inference | TBD | TBD | Yes | LlmInferenceService must reference AITenant |
+| Kuadrant team | TBD | TBD | Maybe | AuthPolicy and RateLimitPolicy per tenant |
+| Platform/OpenShift | TBD | TBD | Maybe | Gateway API, Istio configuration |
+
+**Detailed impacts**:
+
+- **Model Serving**: Core implementation team. Must implement AITenant controller, maas-api multi-tenancy, database schema changes.
+- **Dashboard**: Must add tenant selector, scope all views to selected tenant, handle tenant creation/deletion UI.
```
danielezonca, line 538: I think the exact impact on UI still need to be finalized because overall only GenAI studio is the view that is tenant aware most likely + the tenant creation/deletion UI.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,546 @@
...
+- **Operator**: Must extend maas-controller to reconcile AITenant CRs, manage tenant namespaces.
+- **KServe**: LlmInferenceService CRD must add `gatewayRef` field to reference tenant Gateway.
```
danielezonca: I think this field already exists in LLMInferenceService

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
lphiri: We inject a configmap containing the CA  chain (odh-trusted-ca-bundle )  into namespaces for consumption.   There is is some research to migrate to Trustmanager

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: Correct. We don't touch them because technically they are not owned by a tenant. They are in the same state that they were immediately as they were created. At that point no policy or subscription has been created yet.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: That's fair.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: Although this is mentioned at high level in the impact sections the document stated that the Dashboard approach is a non-goal for this ADR. Do you want me to remove this line ?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: Yes it does. This line was confusing. Removed it.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+  domain: redteam.ai-gateway.apps.example.com
+  
+  # Optional TLS configuration
+  tls:
+    certificateRef:
+      name: redteam-tls-cert
+      namespace: ai-tenants
+  
+  # Optional resource quotas for this tenant
+  resourceQuotas:
```
jland-redhat, line 104: Lets call this a stretch goal for 3.5, I like the idea and I don't see it being to difficult but want to make sure we limits scope

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+  
+  # Resource creation timestamps
+  provisionedAt: "2026-05-27T10:00:00Z"
+```
+
+**MaasTenantConfig CR Structure**:
+
+```yaml
+apiVersion: maas.opendatahub.io/v1alpha1
+kind: MaasTenantConfig
```
jland-redhat, line 130: Talking about this tomorrow but as a note talking to Jim/Dmytro we are going to keep Tenant for EA2 just to make sure we are able to get it out the door since code freeze is coming up.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+spec:
+  # API key defaults
+  apiKeyDefaults:
+    expirationDays: 90
+    autoRotate: false
+  
+  # Observability configuration
+  observability:
+    otelCollectorEndpoint: http://otel-collector:4317
+    metricsPrefix: "maas.redteam"
```
jland-redhat: Is this optional? Can't we just include the tenant is info as part of the metric and limit access based non that? (which I assume is the goal here)

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+    tracesSampleRate: 0.1
```
jland-redhat: I don't think we use this value currently

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+apiVersion: maas.opendatahub.io/v1alpha1
+kind: MaasTenantConfig
+metadata:
+  name: config
+  namespace: ai-tenant-redteam
+spec:
+  # API key defaults
+  apiKeyDefaults:
+    expirationDays: 90
+    autoRotate: false
```
jland-redhat: We don't have AutoRotate functionality so this should be remove or throw a comment it is future functionality

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+    openshift.io/display-name: "Red Team Tenant"
+    openshift.io/description: "To be used only by the red team members."
+spec:
+  # OIDC configuration for this tenant
+  oidc:
+    issuerUrl: https://keycloak.example.com/realms/redteam
+    clientId: ai-tenant-redteam
+  
+  # Domain for this tenant (used for hostname-based routing)
+  domain: redteam.ai-gateway.apps.example.com
```
jland-redhat: Is this something we are going to do for the first iteration? 

I think we just want to let the Gateway determine the domain (for now).

Also the spec needs to add Gateway name.

Can we change this to:

```
gateway:
  name: redteam-gateway
```

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+- Tenant suspend/resume functionality (future enhancement)
+- Per-tenant physical database instances (logical isolation sufficient for v1)
+
+## How
+
+### Architecture Overview
+
+The multi-tenant architecture follows a **one-gateway-per-tenant** pattern. Each tenant gets:
+
+1. An AITenant CR in the `ai-tenants` namespace (cluster-level registry)
```
jland-redhat: We should probably clarify here that our thought is that eventually the AITenant CR would be owned by the "AI Gateway" once we identify exactly what that is.

But we are putting it in MaaS for now since that does not yet exist

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+**Workflow**:
+
+1. Admin deletes the AITenant CR from the `ai-tenants` namespace
+2. maas-controller reconciles deletion and performs the following:
+   - Deletes the Gateway CR (triggers Envoy pod deletion)
+   - Deletes the tenant admin namespace (cascades to MaasTenantConfig, maas-api deployment)
+   - Updates database records: marks tenant as deleted (soft delete)
+   - User namespace CRs (LlmInferenceService, HttpRoute) are **not** deleted
+   - KServe controller updates HttpRoute status to `NotReady` or `Conflicted`
+   - maas-controller updates MaasModelRef status to `GatewayNotFound`
```
jland-redhat: Maybe we should have a clearer status to something like Tenant not found?

Assuming that he MaaSModelRefs -> Tenant which is how they determine Gateways.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
jland-redhat: Also follow up here is this something Kserve would do? Or something that the HTTPRoute would just do automatically when the gateway gets deleted?

I think we either leave the non-existing gateway on the LLMInferenceService or somehow we would have to update it (and I don't think we want to update someone else's object)

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+
+##### Tenant Deletion
+
+**Workflow**:
+
+1. Admin deletes the AITenant CR from the `ai-tenants` namespace
+2. maas-controller reconciles deletion and performs the following:
+   - Deletes the Gateway CR (triggers Envoy pod deletion)
+   - Deletes the tenant admin namespace (cascades to MaasTenantConfig, maas-api deployment)
+   - Updates database records: marks tenant as deleted (soft delete)
```
jland-redhat: To be clear here. The controller has no knowledge of the database. 

We talked about this today if we are going with the Multiple maas-api approach the thought is:


* MaasTenantConfig gets the deletion timestamp (with a finalizer)
* Trigger job against that tenant's maas-api for updating keys
* MaaS Controller tracks job and removes finalizer upon completion 
* MaasTenantConfig + Namespace gets deleted

AI Doc with general approach here:
https://docs.google.com/document/d/12rnvRdakar2R6g_F1pvG_33qB6-3xNHdKjZvXTwF7bk/edit?tab=t.0

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+   - User namespace CRs (LlmInferenceService, HttpRoute) are **not** deleted
+   - KServe controller updates HttpRoute status to `NotReady` or `Conflicted`
+   - maas-controller updates MaasModelRef status to `GatewayNotFound`
+
+**Resource Cleanup**:
+- Tenant namespace: deleted automatically
+- Database api_keys: soft deleted with 90-day retention (configurable)
```
jland-redhat, line 166: Need clarity on what 90-day retention means here, this may be something we target for GA and not TP

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+- User models: remain deployed but unreachable until attached to a new tenant
+- Audit logs: retained per compliance requirements
```
jland-redhat, line 168: Audit logs beyond what is created by Kubernetes around the deletion of the CRs?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+If implemented in future:
+- Suspend would scale Gateway and maas-api to zero replicas
+- Resume would restore to previous replica counts
+- Database records remain active during suspension
+
+##### Default tenant
+
+During a fresh install the following will happen:
+
+1. The controller creates the `nodel-as-a-service` namespace as today
```
jland-redhat, line 188: *model-as-a-service

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+2. The controller creates the AITenant CR in the `ai-tenants` namespace
+```yaml
+apiVersion: maas.opendatahub.io/v1alpha1
+kind: AITenant
+metadata:
+  name: models-as-a-service
+  namespace: ai-tenants
+spec:
+  domain: <existing-domain>  # Preserve existing domain
```
jland-redhat: As noted above I think this should be gateway.name which is an easier thing to look up on.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+```
+3. The controller creates the MaasTenantConfig CR in the `nodel-as-a-service` namespace
```
jland-redhat: *model-as-a-service

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+#### Model Deployment (Model Deployer Role)
+
+**Namespace Flexibility**: Users can deploy models in any namespace. They may use:
+- Separate namespaces per tenant
+- Shared namespace for multiple tenants
+- Per-project or per-team namespaces
+
+**Workflow**:
+
+1. User creates a LlmInferenceService CR and references a single AITenant
```
jland-redhat, line 216: Not sure we have commitment from Model Serving for this for 3.5.

I think that this would still work the way it does today and reference a Gateway which inherently will be attached to a tenant.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+   - **Constraint**: Only one gateway per model (must be an AI Gateway)
+   - **Rationale**: Multi-gateway attachment creates policy management complexity (deferred to future)
+   - **Note**: 1:1 association between tenant and gateway (tenant == ai-gateway)
+
+2. KServe controller creates HttpRoute and InferencePool CRs for the LlmInferenceService
+
+3. KServe deploys the model runtime
+
+4. User creates MaasModelRef CR pointing to the LlmInferenceService
```
jland-redhat, line 225: MaaSModelRef should probably be the one referencing a tenant (note we need a similar flow with ExternalModel as well so I think this makes more sense at the MaaSModelRef level for now)

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+
+**Workflow**:
+
+1. From the tenant admin namespace, create:
+   - MaasAuthPolicy CR
+   - MaasSubscription CR
+
+2. maas-controller reconciles and creates:
+   - Kuadrant AuthPolicy CR
+   - Kuadrant RateLimitPolicy CR (per-user rate limits)
```
jland-redhat: *TokenRateLimitPolicy

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+
+**Now accessible**: Model can be used for inference and is discoverable via `/v1/models` endpoint.
+
+### Data Plane
+
+#### Tenant Identification
+
+Tenants are identified by hostname: `{tenant-name}.{domain}`
```
jland-redhat, line 248: Probably should be via gateway name. We have somewhat limited control over host name.

Also it is possible different tenant have different domains (meaning they could have the same subdomain in theroy.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+#### maas-api Architecture
+
+The maas-api service exposes:
+- Model discovery: `GET /v1/models`
+- API key management: `POST /v1/apikeys`, `GET /v1/apikeys`, `DELETE /v1/apikeys/{id}`
+- Subscription listing: `GET /v1/subscriptions`
+
+**Three Options Evaluated**:
+
+##### Option 1: Dedicated maas-api per Tenant (RECOMMENDED)
```
jland-redhat, line 264: I think we are ok with Option 1 so lets just capture that as the direction

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+- API keys in database without tenant association
+- Models deployed with HttpRoutes attached to global Gateway
+
+#### Upgrade Process
+
+**Phase 1: Create the Default Tenant** (automatic on first reconciliation)
+
+The current models-as-a-service namespace automatically becomes the default tenant. Maas-controller will automatically create the models-as-a-service AITenant CR to make this the actual default tenant. The existent Gateway becomes the gateway for this default tenant. 
+
+1. The controller detects the existent `models-as-a-service` namespace and adds the necessary tenant label.
```
jland-redhat, line 366: Wouldn't step 2 in theory take care of step 1 for us? (Assuming we design AITenant to be able to handle existing namespaces which is probably a good idea anyway)

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,538 @@
...
+2. The controller creates the AITenant CR with `models-as-a-service` name
+3. If there is an existing Tenant CR the controller copies the oidc properties in the AITenant CR. 
+4. The Tenant CR is replaced with MaasTenantConfig CR.
+5. The existent AuthPolicies and MaasSubscriptions will remain unchanged
+4. All LllInferenceServices, MaasModelRef, ExternalModels, ExternalProviders will remain unchanged.
```
jland-redhat: Are we explicitly locking MaaSModelRefs to a tenant I think that is something we talked about.

jland-redhat (commented):
One more aspect we discussed today is currently we determine MaaS health based on the tenant health.


Do we want to base MaaS health on the default tenant health moving forward? Or should it be totally unrelated?

lphiri:
@mariusdanciu @jland-redhat 
We need a way to scope and declare users who can access a tenant. This could be a group list in the AITenant CR or the config object.  This information would be used to create an auth policy to restrict access to the gateway at a layer higher than subcription policy.Think of this as an ingress firewall network policy.
Thoughts?

lphiri:
> One more aspect we discussed today is currently we determine MaaS health based on the tenant health.
> 
> 
> Do we want to base MaaS health on the default tenant health moving forward? Or should it be totally unrelated?

My first thought is we should separate Maas health (and report tenant health on the tenant resource instead)

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: I think it is Istio that updates the status of the HttpRoute when the Gateway is deteled. Kserve does not watch Gateway resources. ( double checked with Claude :) ). I'll update the text.

rhuss:
**Multi-instance orchestration vs. multi-tenancy**

The ADR provides strong admin-side isolation (separate OIDC configs, Envoy pods, maas-api instances, rate limits, database partitioning per tenant). This addresses real operational needs: different identity providers per department, blast radius isolation, independent resource quotas, and per-tenant observability for chargeback. Automating the provisioning of N independent gateway installations via the AITenant CR is genuinely useful.

However, the design does not provide consumer-side isolation. There is no mechanism to restrict which workloads can reach a tenant's gateway endpoint. Any in-cluster consumer with a valid credential (API key or OIDC token) can call any tenant's gateway regardless of which namespace or identity the consumer runs under.

This makes the architecture closer to **multi-instance orchestration** than multi-tenancy in the security sense. The proposal itself is valuable, but the framing matters. Calling it "multi-tenancy" sets expectations around consumer isolation and security boundaries that the current design does not deliver. I would recommend either adding consumer-scoping mechanisms to justify the multi-tenancy label, or reframing as "multi-instance provisioning" to accurately communicate what the ADR provides.

A few questions and suggestions:

1. **Consumer scoping**: Could Kuadrant AuthPolicy enforce audience/scope restrictions tied to the consumer's namespace or service account, beyond just validating a credential?

2. **Network-level isolation**: The ADR doesn't define NetworkPolicy on installation Gateway pods. Should ingress rules restrict which source namespaces can reach an installation's Envoy?

3. **Credential distribution**: API keys are created via MaasSubscription CRs in the installation namespace. How do they reach consumers in other namespaces? Manual copy, Secret reference, ExternalSecret?

4. **In-cluster agent identity**: For workloads deployed via the Kagenti operator (AgentRuntime CRDs), SPIFFE-based identities carry namespace provenance. Could the gateway leverage workload identity rather than just credential possession for in-cluster consumers?

5. **Namespace flexibility**: The current design auto-creates a single `ai-tenant-{name}` namespace per installation. Could the AITenant CR accept an optional list of additional namespaces (e.g., for dev/staging/prod separation within an installation, or for referencing model backends deployed elsewhere)?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: > Do we want to base MaaS health on the default tenant health moving forward? Or should it be totally unrelated?

I don't think the health should be dependent on the default tenant. A client may simply delete the default tenant. IMO The MaaS health should not be affected even if there are zero tenants, because tenants can be created. So maas health should be determined by the controller health.

mariusdanciu: > Is this optional? Can't we just include the tenant is info as part of the metric and limit access based non that? (which I assume is the goal here)

We can use the tenant name as the prefix yes. We can remove the prefix field.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: removed

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: I'd be ok with this but I don't think we should have a separate name for the gateway. I think the tenant name should also be the gateway name, and also the namespace name. We could add suffixes, but I'm not sure what the gain would be.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: No, I think the statement is correct. This is how the tenant is identified in the data plane from the host name string. The Gateway is a control plane object.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: This is already captured by:
```
**Decision**: **Option 1** (Dedicated maas-api per Tenant)
```
and the rationale is below.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: This is an implementation detail. These are logical steps. Yes the implementation could "merge" step 2 and 1 if it always tries to create that tenant. A different implementation might query first and try to create it only if it is missing. This is up to the implementation.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: Depends what you mean by "lock". The MaasModelRef does not need to explicitly point to a tenant because this binding is done by the LlmInferenceService pointing to a Gateway. Once the tenant is gone, the gateway is gone but MaasModelRef is not gone because this is not owned by a tenant. The MaasModelRef is owned by a model deployer which can simply assign it to a different LLMInferenceService or assign the LLMInferenceService to a differen gateway (tenant).

mariusdanciu:
> @mariusdanciu @jland-redhat We need a way to scope and declare users who can access a tenant. This could be a group list in the AITenant CR or the config object. This information would be used to create an auth policy to restrict access to the gateway at a layer higher than subcription policy.Think of this as an ingress firewall network policy. Thoughts?

I thought of that quite a bit in the past few days but today MaaS does not capture users/groups to a scope except in the MaasAuthPolicy (not subscription). This approach makes me wonder if we actually need to capture users in tenants. In SaaS for examples the users are captured by the tenant implicitly because the tenant is the cloud account. It is not the case here. We are essentially stating that a tenant is defined primarily gateway which by definition does know about users. But the question is: what does RHOAI platform wants to do in the future regarding tenants and users ?

Capturing users/groups in the AITenant CR vs the MaasAuthPolicy raises the question on who can manage users for tenants ? If it is the tenant admin then this should not be in the AITenant CR but either remain in the MaasAuthPolicy or MaasTenantConfig CR. Putting users/groups in the MaasTenantConfig CR implies that the auth policy would need to do 2 steps:
- At inference time check if the user belongs to this tenant.
- At inference time check if the user can access model X.

There are other open ended questions here if if user A is part of a MaasAuthPolicy is that user automatically a tenant user OR does it also need to be explicitely added also in the MaasTenantConfig CR ? 

The complexity increases here but we should discuss, yes.

mariusdanciu:
> However, the design does not provide consumer-side isolation. There is no mechanism to restrict which workloads can reach a tenant's gateway endpoint. Any in-cluster consumer with a valid credential (API key or OIDC token) can call any tenant's gateway regardless of which namespace or identity the consumer runs under.

This is correct but I'm not sure how this can be done in any system. Any consumer with the correct URL, credentials **AND authorization policy** can access that tenant for inference. This is true for any system out there and it is also true in SaaS systems.  All tenants/gateways exposes the same type of workloads: inference apis. Please see my notes below on the "In-cluster agent identity" topic. 

> Calling it "multi-tenancy" sets expectations around consumer isolation and security boundaries that the current design does not deliver. I would recommend either adding consumer-scoping mechanisms to justify the multi-tenancy label, or reframing as "multi-instance provisioning" to accurately communicate what the ADR provides.

I'm not sure that I understand the full meaning of consumer-scoping. Are you asserting that tenant X can only be accessed by certain kinds of consumers ? Also what do you mean by consumers ? ... users, groups, types of applications/services ?

> Consumer scoping: Could Kuadrant AuthPolicy enforce audience/scope restrictions tied to the consumer's namespace or service account, beyond just validating a credential?

Since each tenant can be configured with a separate OIDC, this can be achieved. It can even by the same OIDCS but different realm. But this addresses authentication not authorization. Authorization is enforced today via Kuadrant AuthPolicy driven by MaasAuthPolicy,

> Network-level isolation: The ADR doesn't define NetworkPolicy on installation Gateway pods. Should ingress rules restrict which source namespaces can reach an installation's Envoy?

Correct. We do not automate network policies. At this point it is up to the client to apply their own network policies o the pods pertaining to a specific gateway. Also Istio AuthorizationPolicy would be applied for a specific gateway to say allow traffic only from a specific IP block. But this is not automated by MaaS and not in scope of this ADR.

> Credential distribution: API keys are created via MaasSubscription CRs in the installation namespace. How do they reach consumers in other namespaces? Manual copy, Secret reference, ExternalSecret?

No API Keys are not created by MaasSubscriptions. Api keys are created by a REST API exposed by maas-api service and minted for a particular MaasSubscription. Maas "Keys Shown Only Once" Api key is returned only once in the creation response. We only store the hash of the key. So users manage their own apikeys. We don't distribute them.


> In-cluster agent identity: For workloads deployed via the Kagenti operator (AgentRuntime CRDs), SPIFFE-based identities carry namespace provenance. Could the gateway leverage workload identity rather than just credential possession for in-cluster consumers?

I think this is something that can be discussed but the entire idea of "workload identity" should be properly formalized/standardize. Furthermore once the workload identity has be defined and detected, what arre the next steps:
- do we want to capture this in auth-policies ?
- do we want to capture this in rate limiting policies ?
- do we want to capture this in the observability metrics/traces/logs ? 


> Namespace flexibility: The current design auto-creates a single ai-tenant-{name} namespace per installation. Could the AITenant CR accept an optional list of additional namespaces (e.g., for dev/staging/prod separation within an installation, or for referencing model backends deployed elsewhere)?

I don't think this is the case. See, the models themselves are not deployed in the tenant namespace. They can exist in any tenants the user want. However the model is "bound" to a tenant simply because the LllmInferenceService points to an HttpRoute which points to the tenant Gateway. It is the model deployer persona that "assigns" models to tenants.  So the tenant does not technically *own* the model. In the same time nothing prevents a client to allow actual model deployments right in the tenant namespace but the binding is still done in the same way - by the gateway. So there is complete flexibility of where actual deployed models live - as it is the case today for MaaS.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
ishitasequeira: *models-as-a-service

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,530 @@
...
+```yaml
+apiVersion: maas.opendatahub.io/v1alpha1
+kind: AITenant
+metadata:
+  name: models-as-a-service
+  namespace: ai-tenants
+spec:
+
+```
+3. The controller creates the MaasTenantConfig CR in the `nodels-as-a-service` namespace
```
ishitasequeira, line 199: *models-as-a-service

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,530 @@
...
+#### Tenant Lifecycle Management (RHOAI Admin Role)
+
+##### Tenant Creation
+
+**Workflow**:
+
+1. RHOAI admin creates an AITenant CR in the `ai-tenants` namespace
+2. maas-controller reconciles the CR and performs the following:
+   - Creates tenant admin namespace: `ai-tenant-{tenant-name}`
+   - Creates Gateway CR in the tenant namespace
```
ishitasequeira: I noticed that the architecture diagram shows the Gateway CR in the `openshift-ingress` namespace, while this section states that it will be created in the tenant namespace. Could we clarify whether this is an intentional change or if the diagram needs updating?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,530 @@
...
+   - Creates default MaasTenantConfig CR in the tenant namespace
+   - Deploys maas-api service instance for this tenant
```
ishitasequeira: Today, `maas-api` gets installed in the same namespace as `maas-controller` and not the Tenant namespace. Given that we are going with 1 `maas-api` per Tenant, which namespace will the maas-api be installed in? Would it be the Tenant namespace?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
ishitasequeira: +1

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,530 @@
...
+
+##### Tenant Deletion
+
+**Workflow**:
+
+1. Admin deletes the AITenant CR from the `ai-tenants` namespace
+2. maas-controller reconciles deletion and performs the following:
+   - Deletes the Gateway CR (triggers Envoy pod deletion). Consequently Istio updates HttpRoute status to `NotReady` or `Conflicted` when the Gatway is deleted.
+   - Deletes the tenant admin namespace (cascades to MaasTenantConfig, maas-api deployment)
+   - Updates database records: marks tenant as deleted (soft delete).
```
ishitasequeira: *updates api-keys tied to a Tenant as soft deleted (revoked).

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
ishitasequeira: +1, we have hard delete setup only for ephemeral keys and marking keys as `revoked` for other use-cases. There is no cleanup process for revoked api keys.

Also, with soft delete, do we mean setting the key status as `revoked`? Or are we looking to introduce a new deleted status or a soft delete flag?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,530 @@
...
+
+#### Pre-Upgrade State
+
+- Single `models-as-a-service` namespace with one Gateway
+- API keys in database without tenant association
+- Models deployed with HttpRoutes attached to global Gateway
+
+#### Upgrade Process
+
+**Phase 1: Create the Default Tenant** (automatic on first reconciliation)
```
ishitasequeira, line 362: Similar to my earlier comment, today, `maas-api` gets installed in the same namespace as `maas-controller` and not the Tenant namespace. Do we need to add the points around how should we replace existing `maas-api` with Tenant specific `maas-api` instance?

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: The current assumption is that we'll use the same namespace. I added a mention about this in the doc.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: Any maas-api is tenant specific as models-as-a-service is seen as a default tenant. As mentioned above and based on the today call we for now assume that all maas-api instances will continue to live in the same applications namespace. In the future of course can relax that to allow them to live in other namespaces if needed especially if a tenant wants to set a separate DB instance. Currently there is a single 'maas-db-config' secret that must exist in the same namespace with maas-api service.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
mariusdanciu: this has been addressed.

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,557 @@
...
+CREATE INDEX idx_api_keys_tenant ON api_keys(tenant_id);
+
+-- Backfill default tenant for existing keys (migration only)
+UPDATE api_keys SET tenant_id = 'models-as-a-service' WHERE tenant_id IS NULL;
+
+-- Make tenant_id required
+ALTER TABLE api_keys ALTER COLUMN tenant_id SET NOT NULL;
+
+-- Add unique constraint: key hash must be unique within tenant
+-- (allows same key pattern across tenants if needed, though not recommended)
```
jrhyness: @mariusdanciu , since this is not recommended, and we have a unique index on key_hash alone, can we not do this, and require global uniqueness?

andrewballantyne:
You called out stakeholder impacts around Dashboard -- but we were told it's not impacting us in 3.5. We stay fixed to `models-as-a-service` tenancy namespace.

I assume this is holistically impacts us and we should support this, but in the future. Just calling that out here for posterity. @mariusdanciu

architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md
```diff
@@ -0,0 +1,552 @@
...
+- All queries filter by `tenant_id`
+- Unique constraint on `(key_hash, tenant_id)` prevents key reuse across tenants
+
+**Kubernetes RBAC**:
+- Tenant admins have RoleBindings only in their tenant namespace
+- Users cannot list/read other tenants' MaasAuthPolicy or MaasSubscription CRs
+
+**Gateway-level**:
+- HttpRoutes attached only to tenant Gateway
+- No cross-tenant routing possible (separate Envoy pods)
```
lphiri, line 512:
```suggestion
- No cross-tenant routing possible (separate Envoy pods)
- All API requests, including API key generation, gated by authpolicy and subscriptions linked to user identity,
```

lphiri (approved):
Added minor suggestion to make it more explicit that API requests are properly gated to a tenant.  But otherwise LGTM

PR #144, opened 2026-07-01
feat: ai-gateway tenants discovery

mariusdanciu:
## Description

As MaaS introduced multi-tenancy and each tenant comes with its own gateway, any REST API request would have to use the specific Gateway external URL. However so far there was no good way to address discovery the existent tenants in the cluster along with additional information like the gateway URL, namespace etc.

## Merge criteria:

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record covering AI Gateway tenant discovery.
  * Defines a `GET /v1/tenants` discovery endpoint, including how tenant access and gateway connection details are returned.
  * Documents the short-term vs. long-term plan for where discovery is hosted.
  * Describes an in-memory, TTL-style caching approach to reduce request overhead, plus security/privacy considerations and open questions/alternatives.


architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
```diff
@@ -0,0 +1,138 @@
...
+> - This needs to be available even if kserve is not available.
+> - There is not enough time to spin up a new HTTP service to host this endpoint.
+
+Long term, we need a proper discovery HTTP service that exposes the discovery endpoint and potentially, in time, exposes other types of global capabilities that are not scoped per tenant.
+
+### Architecture
+
+![Proposed Architecture](./images/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery-img-1.png)
+
+This proposal introduces a new HTTP Kube service managed by maas-controller that binds to the default Gateway (data-science-gateway).
```
danielezonca, line 68: How does this solution align/overlaps with the similar gateway discovery endpoint built inside odh-model-controller?
- https://github.com/opendatahub-io/odh-model-controller/pull/700

A tenant doesn't entirely overlap with a Gateway instance but there might be the opportunity to consolidate

FYI @pierDipi

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
```diff
@@ -0,0 +1,138 @@
...
+
+- Authn/z - managed via Authorino policy allowing valid OpenShift tokens.
```
danielezonca, line 70: What does it mean in this context?
What is the RBAC requirements that this service will have?

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
```diff
@@ -0,0 +1,138 @@
...
+- The discovery service provides a REST API to query the available tenants in the cluster. The tenant information is maintained in an in-memory cache that is marked as "dirty" after a configurable TTL. This allows limiting the number of kube-api server hits while making the discovery endpoint extremely responsive.
```
danielezonca: Do you expect to have the refresh of this cached list based on time?

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
mariusdanciu: Today we have a gateway discovery endpoint at kserve level IIUC. This API is designed today mainly to exopse the gateway names. This API would not be needed anymore if we introduce the tenants discovery one that advertises the available tenants and the necessary gateway information at maas level. Later on the discovery kube service could be managed by the ai-gateway operator not maas.

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
mariusdanciu: To submit the GET /v1/tenants one needs to have a valid Openshift token. So far the idea is that any openshift user or SA can read the data without specific role bindings to particular RBAC resources. We don't assume a particular type of persona that would require specific RoleBindings. But if you think we do, please let me know.

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
mariusdanciu: That is the simplest approach. There are many ways to refresh the cache and some of them are mentioned in the document in the Alternatives section. We could use a DB table with 1 single row to simply signal that the caches need to be refreshed. Here is a simple flow:

1. maas-controller reconciles AITenants and already reads som Gateway information. 
2. When an update is detected it sends an internal request "cache-is dirty" to any service replica.
3. Cache-dirty flag is set in DB.
4. All replicas will refresh their internal cache.

The DB is just a way to propagate the "dirty" flag to all service replicas. 

TBH I don't think something like this is needed initially. We should try the simples time based approach and test it. I would not expect hundreds or thousands of tenants in a single cluster with frequent changes for tenants.

mariusdanciu: BTW I added `Signal cache refresh via a storage` section in the doc for tracking purposes.

bartoszmajsak: I wonder - can't we just use kube-client and informers with built-in cache to keep things simple?

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
```diff
@@ -0,0 +1,138 @@
...
+##### Tenants cache stored in Postgres
+  The problem with this approach is that while it greatly reduces the number of hits to the kube-api server, it creates more complexity for the service and potentially adds a risk since there is no longer a single source of truth for the tenancy information, but rather two: kube-api and Postgres, and these need to be kept in sync.
+
+##### maas-controller continuously informs the discovery service of any change in the tenants/gateway state
+   The problem with this is the increase in complexity. Also, it is not enough for the maas-controller to send an update request to the discovery service because the service may have N replicas at one point in time, and all replicas would need to be informed about the changes. This can be mitigated with the above alternative of using a DB layer, but again, this would increase complexity. 
+
+
+## Security and Privacy Considerations
+
+None at this point.
```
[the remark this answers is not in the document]
mariusdanciu: The point of the discovery API is for any user to identify what tenants are in the system. Restricting this to platform admins makes no sense as platform admins already can access to the AITenant CRs.

> tenant-membership claims, or filter the response per tenant.
This is a vague statement.

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
mariusdanciu: Technically we can of course but this is what typically controllers do, not so much services. If we have say 2+ POD replicas, each replica will keep a connection open with the kube API server for informers/watch to work. These connections add up on top of many other watchers across the cluster. The idea was to keep this service as lightweight as possible with a minimal footprint on the kube api server. 

If we look at the described alternatives, alternative #2 discusses the maas-controller that already uses the informer pattern and then update the service state via internal REST API. Implementation wise it is a bit more complex than having the service PODs do the watch, but personally I would prefer that since maas-controller already performs the AITenant watch (not yet Gateway watch).

bartoszmajsak: It feels like the lightweight thing to do is to use informers for those 2 CRs to be honest.

The TTL polling approach puts more load on the API server - a `List` call forces serializing the full resource set every time, while a `Watch` is a single long-lived stream that only sends deltas. Using client-go informers is not a controller-specific pattern either - `maas-api` has been doing exactly this since the early days for model listing.

Connection count is negligible in my eyes. We're talking about watching 1-2 resource types (AITenant, maybe Gateway).

jland-redhat: I agree with Bartosz here. Even if we are watching AuthPolicies too that is still relatively small.

And I don't see why would not just use the built in [Authorino cache](https://docs.kuadrant.io/1.0.x/authorino/docs/user-guides/caching/)

lphiri: Can we clarify what we are caching and how this is combined with authorization checks for a specific user or group, given the need to only return gateways specific to a user?

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
```diff
@@ -0,0 +1,158 @@
...
+| Group                         | Key Contacts     | Date       | Impacted? |
+| ----------------------------- | ---------------- | ---------- | --------- |
+| models-as-a-service           | Marius Danciu, Jamie Land | 01/07.2026       | yes |
+| dashboards           | Andrew Ballantyne | 01/07.2026       | yes |
+
+## References
+
+#### REF-1 
+[AI-Gateway multi tenancy ADR](https://github.com/mariusdanciu/architecture-decision-records/blob/main/architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md)
+
```
pierDipi: should it point to the official link? https://github.com/opendatahub-io/architecture-decision-records/blob/main/architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
pierDipi: I tend to agree that the complexity of this approach doesn't seem ideal for this endpoint and considering that caching like this introduces potential user experience problems due to the eventual consistency over that configurable TTL vs a watch that has a negligible delay for an interactive API.

I also have a related question, what's the default TTL we're looking at setting? that directly affects the user experience and the efficiency points brought up before, listing every $TTL might still be less efficient, so is it really worth the complexity + user experience problems?

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
```diff
@@ -0,0 +1,158 @@
...
+    }
+  ]
+}
+```
+The gateway name and namespace are important for model deployer personas, as this information is needed when creating LlmInferenceService CRs.
+
+#### API Scope
+
+The tenants returned by this endpoint should return only the tenants that the authenticated user has access to. This means that if this API returns Tenant A,
+this user or group is part of at least one MaaSAuthPolicy created for Tenant A.
```
jland-redhat: So thinking more about this. Eventually these different tenants will likely have different Authentication setups and different IDPs

I don't know we can make the assumption that an Openshift user the same user info as the OIDC provider. Or even that our users will be Openshift users at all.

mariusdanciu:
@danielezonca @lphiri please revise `Cache hydration` section. Base on slack conversations using informers is preferred to keep the cache in sync. I am not a huge fan of this approach but provided that this API does not involve large volumes of data or very frequent changes for AITenant or Gateway CRs this is not a major concern at this point. But development wise this option seems to be the easiest one.

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
mariusdanciu: That's a good point since the discovery sits at a higher level than tenants it may one the openshift token or the OIDC setup at openshift level. This may be different than the OIDC configured at tenants level. I'm OK with relaxing this assumption to "show all available" tenants to all authenticated users but this again assumes that the calling users would have openshift tokens. If that is an impediment than that discovery should be public requiring no authentication. This would be similar with OIDC discovery i.e. https://accounts.google.com/.well-known/openid-configuration. @lphiri any thoughts ?

lphiri: I think we need an ADR just for this part  : )  What is the use case for tenant discovery? If its for the RHOAI UI Dashboard, then with the current implementation  users are going to call the API in the context of an Openshift console session and hence will have an Openshift/k8s token. 
We can return all available gateways without authentication/filtering, but I am not sure that is useful if the user does not have access to all the gateways returned.

mariusdanciu: Well this is the ADR for tenants discovery. There is a current Gateway listing in Kserve component but we need tenants discovery regardless if Kserve is there or not. End-users needs to know what is the URL for a specific tenant so they can run inference and model deployer persona need to know the Gateway info in order to create LllmInferenceService or ExternalModels CRs. 

Returning all available tenants in the system implies that user A would also see tenant that it doesn't have access too.  I see no security risk in that. Regarding authentication if user A has access to tenant T and tenant T is configured with a 3rd party OIDC, it doesn't necessary mean that user A also exist in Openshift so it can use the Openshift token. This is why there is the question if the discovery endpoint should be exposed without authentication similarly with how OIDC discovery is. If is make authentication mandatory then user A must exist in both Openshift and the Tenant's OIDC.

lphiri: @mariusdanciu  the aspects I think we need to be more specific is how we expect the users of the discovery API to use it.  My concern is not so much the security aspect, but rather how useful the returned data is. 
For example, if I am a user attempting to set up an LLMInference service,  how do I know which Gateway in the returned unfiltered list I am allowed to attach to without having to try each one? The KServe implementation filters the returned list to address this issue.
On the Openshift user aspect,  again this depends on whether we expect users in the RHOAI dashboard to use the discovery service, if so, these users must have Openshift/k8s tokens,  because they need to have namespace access to create LLMInference services.

mariusdanciu: > For example, if I am a user attempting to set up an LLMInference service, how do I know which Gateway in the returned unfiltered list I am allowed to attach to without having to try each one?

This is exactly what this ADR address. The questions is not about how to access LLMInference directly but rather how to access it via the AIGateway. KServe is not mandatory as MaaS can still work with ExternalModels. Therefore having the gateway discovery at kserve level is problematic and this is why we need a higher level discovery service for AITenants not just for gateway names, but for ai tenants (In time we may need to expose more metadata). The exact payload of this API is described in this ADR and yes of course this also contain the gateway information as well as the tenant information. Example:

GET /v1/tenants

```json
{
  "tenants": [
    {
      "name": "models-as-a-service",
      "namespace": "models-as-a-service" 
      
      "gateway": {
         "name": "maas-default-gateway",
         "namespace": "openshift-ingress",
         "protocol": "https",
         "externalUrl": "https://maas.apps.rosa.qzhq9-eu9g5-5aj.orkd.p3.openshiftapps.com",
         "port": 443,
      }
    }
  ]
}

```

This is how users and the Dashboard know what to call for specifci tenants. 

> On the Openshift user aspect, again this depends on whether we expect users in the RHOAI dashboard to use the discovery service, if so, these users must have Openshift/k8s tokens, because they need to have namespace access to create LLMInference services.


Correct the discussion that we had with @jland-redhat above goes a little beyond that. Lets say that an openshift user authenticates and uses the Dashboard. Now let's say that tenant T is configured with a different OIDC and user U1 exists in that OIDC. But U1 does not exist in openshift. This means that U1 cannot use the discovery endpoint to find out the tenant info above. This is not about a regular Dashboard flow, but rather about API usage. If our position is that you have to have an openshift user to use the discovery, I'm fine with that but we need to document the above case. This is why a no-auth access to the discovery (read only data) would make sense to me, for the same reasons OIDC .well-known configuration are exposed with no-auth. But you are the security expert so I'm ok adopting your recommendation here. The ADR as is, assumes valid openshift tokens to access the discovery endpoint. 

P.S. Yes Dashboard users would be openshift users. 

Looking forward for your feedback !

lphiri: My usability concerns were based on the assumption that the user starts with zero bootstrap info, which would make it hard for them to choose the right option from the returned data.
If we assume the user already has basic context—like the tenant name—and only uses the discovery service to fetch additional gateway details, those UX concerns go away (and the OIDC analogy makes much more sense).
With that clarified, I'm fine moving forward with an unauthenticated discovery service. My only remaining request is that we avoid returning the tenant's namespace, as some users might consider that sensitive information (only tenant admins really need to know this anyway).

mariusdanciu: @lphiri I think we clarified this on the call. I updated the ADR with 2 main points:
1. Removed the ai-tenant namespace fields from the response payload as historically customers raised questions on namespaces exposure.
2. Clarifies the API scope to not require authentication.

If this looks good to you and there are not other questions, hopefully we can approve the ADR.

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
```diff
@@ -0,0 +1,161 @@
...
+    }
+  ]
+}
+```
+The gateway name and namespace are important for model deployer personas, as this information is needed when creating LlmInferenceService CRs.
+
+#### API Scope
+
+As this endpoint exposes the read only list of available tenants in the system we don't require authentication. This is similar with how OIDC .well-known configuration API is exposed. Also, this allows users that exist in external OIDC systems but not in Openshift to use the discovery API.
+
```
lphiri:
```suggestion
It is assumed the API consumer already possesses the target tenant name and is querying for additional metadata.
```

architecture-decision-records/model-serving/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery.md
```diff
@@ -0,0 +1,161 @@
...
+| Group                         | Key Contacts     | Date       | Impacted? |
+| ----------------------------- | ---------------- | ---------- | --------- |
+| models-as-a-service           | Marius Danciu, Jamie Land | 01/07.2026       | yes |
+| dashboards           | Andrew Ballantyne | 01/07.2026       | yes |
+
+## References
+
+#### REF-1 
+[AI-Gateway multi tenancy ADR](https://github.com/mariusdanciu/architecture-decision-records/blob/main/architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md)
```
lphiri:
```suggestion
[AI-Gateway multi tenancy ADR](https://github.com/opendatahub-io/architecture-decision-records/blob/main/architecture-decision-records/model-serving/ODH-ADR-MS-0003-ai-gateway-tenancy.md)
```

lphiri (approved):
@mariusdanciu  a couple of suggested updates, but ok with approving.

PR #146, opened 2026-07-10
add ADR decouple cert-manager from cloud controller manager

davidebianchi:
## Description

- Add ADR proposing to move cert-manager installation from the cloud controller manager to the XKS Helm chart
- Cloud controller manager retains PKI management (ClusterIssuers, root CA, webhook certificates) but no longer deploys cert-manager itself
- Defines migration strategy: chart blocks upgrade via `lookup` if live KE CR still has `managementPolicy: Managed`, guiding user through two-step migration

Jira task: https://issues.redhat.com/browse/RHOAIENG-69977

## How Has This Been Tested?

ADR document — no code changes. Cross-referenced with cloud controller manager code in `opendatahub-operator` to verify claims about cleanup behavior, GC protection, ownerReferences, and cert-manager-operator finalizer lifecycle.

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record clarifying cert-manager installation ownership versus PKI lifecycle management.
  * Documented how to control cert-manager installation through chart options, including default behavior and migration steps for existing installs.
  * Added upgrade-time validation guidance to prevent disruptive changes, including expected downtime and which resources are preserved.

### DIFF
--- a/architecture-decision-records/operator/ODH-ADR-Operator-0014-decouple-cert-manager-installation.md
+++ b/architecture-decision-records/operator/ODH-ADR-Operator-0014-decouple-cert-manager-installation.md
@@ -33,7 +33,7 @@ Several factors make cert-manager a poor fit for controller-managed installation
 * Establish a clear boundary: the deployment mechanism installs cert-manager, the CCM manages PKI resources (ClusterIssuers, root CA)
 * Support environments where cert-manager is already installed
 * Define a migration path for existing deployments where the CCM currently owns cert-manager
-* Remove the `certManager.managementPolicy` field entirely from the CCM CR API
+* Deprecate the `certManager.managementPolicy` field from the CCM CR API
 
 ## Non-Goals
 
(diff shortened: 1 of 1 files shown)


architecture-decision-records/operator/ODH-ADR-Operator-0013-decouple-cert-manager-installation.md
```diff
@@ -0,0 +1,173 @@
...
+* When `certManager.enabled: false`, the chart assumes cert-manager is already present in the cluster and skips installation
+* The chart will pin a specific cert-manager version to ensure compatibility with the controller's PKI expectations
+
+### Controller changes
+
+The controller will:
+
+* Remove all cert-manager installation logic (CRD deployment, namespace creation, Deployment management)
+* Retain all PKI resource management (ClusterIssuers, root CA Certificate, Secrets)
+* Fail with a clear error if cert-manager CRDs are not present at startup, rather than attempting to install them
```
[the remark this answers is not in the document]
davidebianchi: This is already managed in cloud controller manager: the certificates will be eventually installed in the cluster once CRD exists and cert-manager webhooks are ready, otherwise the xKE CR will raise an error

architecture-decision-records/operator/ODH-ADR-Operator-0014-decouple-cert-manager-installation.md
```diff
@@ -0,0 +1,152 @@
...
+| cert-manager lifecycle (upgrades, scaling) | Deployment mechanism | cert-manager Deployment, CRDs |
+| PKI resources (root CA, ClusterIssuers, per-namespace and webhook certificates) | CCM | ClusterIssuer, Certificate, Secret |
+| GC protection of PKI resources | CCM | GC-protected ClusterIssuers and root CA Certificate |
+| CertManager monitoring condition | CCM | Status conditions on CCM CR |
+
+### Configuration
+
+After decoupling, cert-manager installation is no longer controlled by the CCM.
+
+The `certManager.managementPolicy` field is removed entirely from the CCM CR API. The `certManager` field itself is retained for PKI-related configuration (e.g., certificate names, issuer references). Designing the PKI configuration surface is out of scope for this ADR.
```
asmigala: Will removing the field completely break backwards compatibility? Should we perhaps leave it but enforce it being set to `Unmanaged` at runtime? I recall we did something similar for DSC before

architecture-decision-records/operator/ODH-ADR-Operator-0014-decouple-cert-manager-installation.md
```diff
@@ -0,0 +1,152 @@
...
+* **Root CA Secret** (in `cert-manager` namespace): created by cert-manager in response to the Certificate CR. Has no ownerReference to the CCM CR, so it is not affected by cleanup.
+* **ClusterIssuers**: cluster-scoped, protected as long-lived PKI infrastructure by the GC mechanism
+
+### Boundary definition
+
+The responsibility split between installation mechanism and controller after decoupling:
+
+| Responsibility | Owner | Resources |
+| --- | --- | --- |
+| cert-manager installation (CRDs, Deployment, RBAC) | Deployment mechanism (e.g., Helm chart, manual) | cert-manager namespace, Deployment, CRDs |
```
asmigala, line 71: Do I understand correctly that we currently deploy the `cert-manager-operator` which in turn deploys the `cert-manager` itself, while in this new architecture, we are going to deploy `cert-manager` directly using the helm chart?

Are we going to use the upstream helm chart as a dependency of the odh-gitops for that? Or leave the installation completely to the user?

davidebianchi: No, we'll deploy from the helm chart the cert-manager-operator, which deploy the cert-manager. But the user could always decide to install directly the cert-manager (or it could already be installed in the cluster).

The real change is the actor who install the cert-manager: from cloud controller manager to the user

architecture-decision-records/operator/ODH-ADR-Operator-0014-decouple-cert-manager-installation.md
davidebianchi: Not sure, to verify if additional fields are allowed. In case it breaks the backward compatibility yes, we should keep the deprecated field until the version bump. I'll specify it here. From the architectural point of view, the important think is that the logic to install the cert-manager-operator is removed from the cloud controller manager

architecture-decision-records/operator/ODH-ADR-Operator-0014-decouple-cert-manager-installation.md
asmigala: Do you think this distinction is worth capturing in the ADR?

architecture-decision-records/operator/ODH-ADR-Operator-0014-decouple-cert-manager-installation.md
davidebianchi: Updated, can you take a look?

architecture-decision-records/operator/ODH-ADR-Operator-0014-decouple-cert-manager-installation.md
davidebianchi: Updated, can you take a look?

asmigala (approved):
LGTM

lphiri:
@davidebianchi  this generally looks good to me.   Can we also run this by the other stakeholders/components that dependent on Cert Manager?

davidebianchi:
Thanks @lphiri! In xks, the cert-manager should be created externally and the platform certificates are created by the cloud controller manager, with the opendatahub-operator which pass the configuration needed to the component module. So, the component can use the platform ca etc.
 Do you have a specific list of components in mind?

lphiri:
> Thanks @lphiri! In xks, the cert-manager should be created externally and the platform certificates are created by the cloud controller manager, with the opendatahub-operator which pass the configuration needed to the component module. So, the component can use the platform ca etc. Do you have a specific list of components in mind?

@davidebianchi  actually ignore my comment -  everyone I had in mind should be included in RHAII.

davidebianchi:
Ok, is this ADR ok to be merged?

lphiri (approved):
LGTM

PR #149, opened 2026-07-21
Data connect hub

mariusdanciu:
## Description
We are introducing the DataConnectHub service and operator to:
- allow secure data ingestion for clients that need to read data from multiple data sources without maintaining a potentially large number of drivers and 3rd party lib dependencies
- Data connections management 

## Merge criteria:

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record for Data Connect Hub, outlining ingestion patterns, DataConnection access model, multi-tenancy (including soft-tenancy via tenant header), APIs, and security/observability expectations.
  * Updated Data Connect Hub diagrams to expand from 4 to 5 pages, including a new “Secret sync” flow and clearer hard- vs soft-tenancy views.
  * Added an architecture decision record for AI Gateway tenants discovery, including the `GET /v1/tenants` behavior with permission-scoped responses and caching via informer-style watches.


architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+  - DataConnectService CR is created by the tenant in the tenant namespace. This triggers the creation of the DCH service instance in the DCH infra namespace
+  - Implies the existence of a tenants discovery service since each tenant implies a different host name
+  - Can support separate metadata store per tenant.
+
+- **Soft tenancy** - There is a single DataConnectService CR in the opendatahub namespace and this is managed by the cluster admin not by the tenant admins. This implies:
+  - There is a single DCH service instance (that scales horizontally of course)
+  - There is a single HttpRoute and a single Gateway. 
+  - There is a single Praxis instance for all tenants.
+  - All tenants use the same hostname / URL
+  - Tenants can no longer be distinguished by the hostname. Instead we will require the presence of the `x-tenant-id` header. This is similar to EvalHub and MLFlow approaches for multi-tenancy.
```
etirelli, line 96: please add requirement that user auth is checked for access to `x-tenant-id` (e.g. SSAR check).

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+    - admin properties (never shared with end users)
+        - url + port
+        - credentials secret_ref
+    - user visible properties
+        - name
+        - description
+        - data source name
+        - ...
+- `ConnectionSubscription` - defines which users/groups can access what Connections with what limits.
+    - subjects list
```
etirelli: Looking at the example CRs below, it seems to me that this ADR is reimplementing the authorization framework. 
Wouldn't it be better to have the ConnectionSubscription to reference a Role instead and rely on RoleBindings to give access to subjects? By using Roles and RoleBindings:
1. you wouldn't need to re-implement the auth framework. Admins already have proper workflows and tooling to work with Roles and RoleBindings.
2. you get finer granularity on permission assignment via verbs (get, list, read-data, etc) instead of binary has access to/doesn't have access to as per the current spec.
3. in the current spec, each change on access requires changes to the ConnectionSubscription CR. Imagine 10 connections, with 10 subscriptions each (for different limits), and a new user added to the system would require 100 CRs to be updated.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+
+```yaml
+apiVersion: dataconnect.opendatahub.io/v1alpha1
+kind: ConnectionSubscription
+metadata:
+  name: evals
+  namespace: ai_trust
+spec:
+  description: Subscription for evals
+  subjects:
```
etirelli: see my previous comment about defining subjects inline in the CR.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+  ## Optional
+  requests_limits:
+    - limit: 30
+      window: 1m
+
+```
+
+**Notes**
+- If a DataConnection is not attached to a ConnectionSubscription, it is not usable for data reading.
+- If the subjects contains a user with name `*` this is an explicit statement that the connection referenced by this subscription is **PUBLIC** and any user can access it. Thus any tenant can define public connections.
```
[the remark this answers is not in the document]
etirelli: +1, need to define the proper semantics of PUBLIC ("*"). It should require at minimum that the user is authenticated to allow access.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+* Define the tenancy model
+* Define access management
+
+
+## Non-Goals
+
+* OTLP
+* ETL
+* Document processing
+* Higher level governance aspects (lineage, catalog, etc)
```
abiazett, line 39: Please add - management of data schemas, as this is a concern of the Data catalog/registries

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+
+#### Data Ingestion
+
+DCH service proposes data ingestion APIs via arrow flight, or HTTP API. This ensures that the client using this API (or SDK) does not need to install and maintain various DB drivers or 3rd party libraries and this also implies that the actual database credentials are never exposed to these clients. However, while there are use cases where such API abstraction is very useful, in other contexts using the 3rd party libraries is preferable. DCH is not opinionated on the approach that clients want to adopt. However, there are implications that require mentioning here. If a client needs to use 3rd party libraries to connect directly to various data stores, it means that these clients require the actual backend credentials to connect. DCH service APIs are not used. However, the DCH Operator that watches DataConnection CRs will automatically create a new kube secret (if the DataConnection CR is configured as such) with the name `{dataconnection CR name}-secret` and this secret entails the DataConnection CR metadata + the actual credentials secret information that the DataConnection CR is pointing to. Visually this looks like:
+
```
etirelli, line 306: missing diagram?

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+```
+  DataConnection CR + credentials secret = New secret
+```
+
+DCH Operator will ensure that the new secret will only be accessible by the subjects specified in the ConnectionSubscription CR. In other words, it will create and maintain the Role and RoleBinding for this new secret. In other words, if an application (i.e. a notebook) wants to mount this new secret, it needs to use a user or a ServiceAccount that is specified in the ConnectionSubscription CR. Without these, the Role and RoleBindings will not exist.
```
etirelli: if the DCH operator will maintain Role/RoleBindings based on the ConnectionSubscription, why duplicate the list of subjects inside the ConnectionSubscription CR?

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+- DCH Service management as described in this doc
+- REST APIs for end-users connections listing.
+- Arrow-Flight APIs for tabular data reading.
+- REST APIs for unstructured data reading.
+- Access control implemented in a sidecar container - no Kuadrant dependencies yet. This does not have to be throwaway code because this can be very useful for upstream adoption when Kuadrant is not available. 
+- A few tabular data sources, tabular data formats (parquet, csv, json) and object store data sources support (to be agreed with PM)
+
+## Open Questions
+
+TBD
```
etirelli, line 417: What is the plan for the Data Connection implementation that currently exists in the Dashboard? Can they be migrated to this solution?

architecture-decision-records/data-connect-hub/Untitled
```diff
@@ -0,0 +1 @@
+![Proposed Architecture](./images/ODH-ADR-MS-0004-ai-gateway-tenancy-discovery-img-1.png)
```
etirelli: please update file name

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+  - Tenants can no longer be distinguished by the hostname. Instead we will require the presence of the `x-tenant-id` header. This is similar to EvalHub and MLFlow approaches for multi-tenancy.
+  - Tenant namespaces need to be labeled with `dch.opendatahub.io/tenant` so that DCH Operator knows which namespaces to monitor. These are the namespaces where the `DataConnection`, `ConnectionSubscription` CRs are managed.
+  - Tenants are segregated by a tenant-id field in the metadata store (Postgres)
+
+
+- **Tenant admin persona** - has access to the tenant namespace and manages `DataConnection`, `ConnectionSubscription` and `DataConnectService` (for hard tenancy case) CRs. These CRs are described later on in this document.
+- **Data consumer persona** - this can list available connections that this identity has access to and perform data ingestion via provided APIs, SDK or CLI.
+
+##### Note
+For the initial release we will adopt the soft tenancy approach as this implies less development and does not require a separate tenants discovery service.
```
abiazett, line 101: Could you please clarify how other components, like Data Registry which is name spaced will find the right  DataConnectService and its Dataconnections in order to show as options for when linking a DataRegistry DataSet with a Dataconnection?

etirelli (dismissed):
See previous comments.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
abiazett: I agree with @etirelli , Data Connector/Registry uses RHAI RBAC, referencing Roles and Role Bindings, similar to MLFlow, and we would like to keep the overall admin experience similar across components in the platform.

abiazett: @mariusdanciu btw, with K8/RHAI RBAC we are able to do access control at the "instance" level, here is an example: 
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: sarah-specific-tables
  namespace: ml-project-alpha
rules:
- apiGroups: ["datacatalog.opendatahub.io"]
  resources: ["tables"]
  resourceNames: ["document_registry"]    # ONLY this table
  verbs: ["get"]
# sarah can GET document_registry but NOT other tables in the namespace

more info here: https://gitlab.cee.redhat.com/data-strategy/data-arch-design-options/-/blob/main/docs/adr/internal-0002-catalog-api-rbac.md?ref_type=heads

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
abiazett: I am not clear on how this option is helping the scenario where the user does not use the DCH service APIs? The user still needs to create a connectionSubscription CR? Please describe the complete scenario. Also, we want to make sure that the "external access" using Iceberg Table format using existent libraries is possible.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+metadata:
+  name: my_eval
+  namespace: ai_trust
+spec:
+  description: Connection to the test data S3 bucket
+  type: s3
+  format: jsonl
+  properties:
+    size: 200MB
+  schema:
```
abiazett: @mariusdanciu as discussed on the arch review with Data Catalog as part of Data Strategy, schema of a dataset belong on the Data catalog/Registry Table/Volume object. Could you please remove it from here as we don't want to duplicate this info. As we discussed Apache Arrow may use schema info dynamically as part of the ingestion, but it should not be saved on the connection CR. In follow on phases, as we look at schema discovery/inference, we would get the info discovered (potentially through DCH) and store it into the Data Catalog/Registry.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,356 @@
...
+  subjects:
+    # Users, groups or service-accounts can be specified here. 
+  - kind: user
+    name: jim
+  - kind: user
+    name: karen
+  connections:
+    - name: my_eval
+  ## Optional
+  requests_limits:
```
abiazett: I thought we had agreed on no rate limiting for phase 1, in order to keep a small scope we can deliver, get feedback from clients and then expand.

abiazett (dismissed):
@mariusdanciu I see a couple of inconsistencies on what I thought we had agreed upon on the meeting of 07/20: https://docs.google.com/document/d/1Q66nd6ysuBT7ZgSxVzUrRU7hUZStQ08YL7UFTo5buCU/edit?tab=t.8x80kzkogzh4. Please let me know if you don't agree, in which case we need to reconnect for impact on the Data Catalog/Registry side.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: 1. DataCatalog does not really need to know about DataConnectService CR since this is about the actual service instance management. For the soft-tenancy (MVP) approach as the the diagram, the DataConnectService CR lives in the same namespace as the data-connect-hub-operator (i.e. opendatahub) and it is managed by the cluster admin.

2. From DataCatalog you can query the DataConnection CRs from the all the namespaces that the user has access to. For listing/discovery purposes only I would recommend DataCatalog to use the REST API for DataConnection listing as this will return the DataConnections that this user has access to based on the Subscriptions defined. But it would be better to let us know what types of queries DataCatalog will need from DCH.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: No, the user does not create ConnectionSubscription. The tenant admin does. Also DataConnection CR are assumed to be managed by the tenant admins not end users. 

There is a slight misunderstanding about ConnectionSubscription. This is a high level object from this the operator creates:
- Authorino policies
- Limitador RateLimits policies
- Kube Role and RoleBinding for the specific DataConnection CR. I will make this item more explicit in the doc.

But for external data consumption you will need to use the new-secret created by the operator as you need secret mounting capabilities, not the DataConnection CR itself. The secret itself as described in the doc is protected by RABC.

So the flow is:

1. Tenant admin creates the DataConnection CR
2. Tenant admin created the ConnectionSubscription CR
3. DCH operator creates the AuthPolicy, RateLimitPolicy, Role and RoleBindings for this DataConnection CR 
4. DCH operator create new new-secret (DataConnection CR + credential secret) and the Role and RoleBinding for the secret itself. This secret is also label pointing to its DataConnection CR

DataCatalog can query the new-secret based on the label and mount it in other PODs (i.e notebooks etc). The secret is ready to be mounted.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Yes I thought I removed it but apparently not. Will do. However while schema discovery does belong to DataCatalog this does not mean that DCH is unaware of schema. This is impossible because ArrowSQLFlight itself needs to provide the schema to the client. It is how the protocol works.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: This document is not only about MVP. This is an architecture document for DCH not a planning doc. The rate-limits need to remain in the doc. Phasing the deliverables is done with PM based on RFEs.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: @etirelli We are most definitely not re-implementing the authorization framework. ConectionSubscription CR is a high level object that contains descriptive information. Based on this the operator creates the Authorino policies, Limitador policies and RBAC Roles/Roles binding.  The authentication and autheorization is handles by already existent platform components: Authorino (for REST gRPC flows) and RBAC for direct CRs secrets access.

> you get finer granularity on permission assignment via verbs (get, list, read-data, etc) instead of binary has access to/doesn't have access to as per the current spec.

End users only need to read data connection (well some information). DataConnections are managed only at kube level by tenant admins. The TenantAdminRole will be automatically created by the operator.

> in the current spec, each change on access requires changes to the ConnectionSubscription CR. Imagine 10 connections, with 10 subscriptions each (for different limits), and a new user added to the system would require 100 CRs to be updated.

I don't understand your argument. Why ? Adding a new user to a DataConnection would imply simply adding it to the ConnectionSubscription. Also a subscription can point to multiple DataConnecation. So with your example a single subscription can point to all 10 DataConnections and adding a user to a Subsription is literally a single change. If for some reasons they want 10 different limits yes different subscriptions are needed. But this is up to the client how they want to administer their subscriptions. MaaSubscriptions works exactly like this as well. In practice there is no point in defining lots of different rate-limits.  

This pattern is already used in MaaS and AI-Gateway. 

I added the section `What happens under the hood` inside the access control. So again no we are not creating a new auth-framework. There is not a single component in DCH that performs authentication and authorization in the (long term). For MVP we may not need Authorino and Limitador but this is a planning discussion. The point is that ConnectionSubscription is a descriptive object and all the necessary bindings are done by the operator. Otherwise relaying only on Roles and RoleBindings will create a lot of manual work and another aspect is that we don't want to hit kube api server to perform SAR for data ingestion APIs which can be frequent. In time we may need to adopt separate api-keys for authentication as maas did not we're not there yet. Hopefully some of these things can exist at platform level so we can re-use the,

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: I responded.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: TokenReview check is not possible when each tenant uses a separate OIDC. x-tenant-id identifies the namespace of the scope of the query. The request will fail if the subject is not part of any subscription for this tenant. Note that this flow is about the DCH ingestion APIs (via REST or Flight Arrow) not via Kube APis.

The ConnectionSubscription goal is to define access for DCH ingestion APIs that need to specify a Connection where to ingest from. It is not meant to replace RBAC. A tenant admin can still define whatever roles and role bindings they want for the DataConnection CR themselves. Thus SAR checks happens whenever a subject wants to access the CRs objects via Kube API. 

I will make these more explicit in the ADR.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: No. Visually it looks like:

```
  DataConnection CR + credentials secret = New secret
```

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Because the ConnectionSubscription is the high level CR that admins manage. This triggers under the hood access setup for Authorino, Limitador and RBAC.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Yes at some point we'll need to migrate them, but for now we are not touching it.

architecture-decision-records/data-connect-hub/Untitled
mariusdanciu: That was a rogue file. Thanks for pointing it.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Done.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
etirelli: > We are most definitely not re-implementing the authorization framework. 

Sorry, what I meant is that by defining user-authorization inside ConnectionSubscription you are creating an abstraction layer on top of familiar existing APIs. It prevents admins from using their tools to manage access and force them to use a completely new API. It creates additional complexity when managing users at scale (give a new user access to multiple connections, remove access of a user from multiple connections, audit which connections users have access to, etc).

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
etirelli: closing this comment, lets consolidate the discussion on the previous one.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
etirelli: It would be good to acknowledge the overlap in this ADR, and propose the plan to deal with it during the transition. For instance, how do we make clear for users/customers when to use one vs when to use the other?

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: This is intentional because admins have a single place to go rather that manage: Roles&RoleBindings, Authorino policies, Limitador policies. I don't really agree that it creates additional complexity as a higher level construct makes things easier. Same applies for MaaS.

> (give a new user access to multiple connections, remove access of a user from multiple connections, audit which connections users have access to, etc).
Removing a user access is simply removing the user from the subscription and the user would no longe have access to any DataConnection (can be many) referenced by this subscription.  So working with a higher level CR actually simplifies the admin work rather that dealing with at least 4 types of lower level CRs.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: That's a good point. I will add a section to cover this. Thanks !

mariusdanciu:
@etirelli @lphiri @danielezonca I added a dedicated section for `ConnectionSubscription CR` and also a separate one for `From RHOAI Connections to DCH DataConnections`

lphiri:
@mariusdanciu  - this MR seems to include two unrelated PRs - having trouble reviewing the changes from the diff:
https://github.com/opendatahub-io/architecture-decision-records/pull/149/changes

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,442 @@
...
+#### ConnectionSubscription CR
+
+Defines the data access rules when using the DCH ingestion APIs. It does not define how the access to the DataConnection CR is managed. Therefore admins can still define their own RBAC rules for access to the DataConnection CR objects.
+
+**What happens under the hood ?**
+- Tenant admin creates the DataConnection CR
+- Tenant admin creates the ConnectionSubscription CR
+- DCHO (DataConnectHubOperator) creates the Authorino AuthPolicy and Limitador RateLimit policy. This is needed for REST and Arrow Flight API access. (This is for the long term)
+- DCHO creates the Role and RoleBinding for data ingestion pointing to a DataConnection CR by the subjects mentioned in the ConnectionSubscription.
+
```
lphiri, line 265: What happens if this RoleBindings are manually created?

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,442 @@
...
+- `REST` API for:
+    - reading/listing Connections information
+    - chunked or full data reading for unstructured datasets.
+    - small sample data reading mainly for UI rendering and connection test purposes.
+- `Arrow-Flight` - a gRPC protocol for efficient tabular data reading in columnar format. This is suitable for SQL datastores but also for vector databases. 
+- `CR` APIs - for tenant admin management.
+
+### CR Admin APIs
+
+- `DataConnectService` - describes the DCH service instance and configurations for this tenant. OIDC, hardware resources etc.
```
lphiri: OIDC configuration should only be done once per tenant, not repeated for each service.    There is another opportunity to consolidate here.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,442 @@
...
+  - DataConnectService CR is created by the tenant in the tenant namespace. This triggers the creation of the DCH service instance in the DCH infra namespace
+  - Implies the existence of a tenants discovery service since each tenant implies a different host name
+  - Can support separate metadata store per tenant.
+
+- **Soft tenancy** - There is a single DataConnectService CR in the opendatahub namespace and this is managed by the cluster admin not by the tenant admins. This implies:
+  - There is a single DCH service instance (that scales horizontally of course)
+  - There is a single HttpRoute and a single Gateway. 
+  - There is a single Praxis instance for all tenants.
+  - All tenants use the same hostname / URL
+  - Tenants can no longer be distinguished by the hostname. Instead we will require the presence of the `x-tenant-id` header. This is similar to EvalHub and MLFlow approaches for multi-tenancy.
```
lphiri, line 96: I dont follow how we plan to  use a single HttpRoute/URL.
Can you expand on how this would work?  How will the routing logic uniquely identify the target connection ? If we are using path based routing can we make that explicit? I undertand how 'x-tenant-id' can be used to distinguish tenants,  but not specific connection. The standard way to this is include namespace in the url path.

On the tenant-id part , I understand  Praxis has in built multi-tenancy support at the gateway level  -  I dont know the details but that may also be worth considering here.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,442 @@
...
+##### Soft tenancy
+
+![Fig 3](./images/ADR-0001-img4.png)
+
+- **Hard-tenancy** - Each tenant has its own DCH service instance that points to a dedicated gateway. This means that the Praxis proxy is also dedicated. This means:
+  - Each tenant gets its own hostname / url
+  - DCH service instance is dedicated to that tenant
+  - Traffic is isolated per tenant (no neighbour noise)
+  - Different TLS certificates managed via cert-manager can be used per tenant.
+  - DataConnectService CR is created by the tenant in the tenant namespace. This triggers the creation of the DCH service instance in the DCH infra namespace
```
lphiri: Gateways are generally considered "expensive"  infrastructure components.  
Can we plan for Gateway consolidation at the ODH level from the start in this mode (single gateway per tenant , instead of one gateway per tenant service)? This would mean potentially reusing MaaS gateway for DCH.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,442 @@
...
+
+```
+  DataConnection CR + credentials secret = New secret
+```
+
+DCH Operator will ensure that the new secret will only be accessible by the subjects specified in the ConnectionSubscription CR. In other words, it will create and maintain the Role and RoleBinding for this new secret. In other words, if an application (e.g. a notebook) wants to mount this new secret, it needs to use a user or a ServiceAccount that is specified in the ConnectionSubscription CR. Without these, the Role and RoleBindings will not exist.
```
lphiri: Probably missing some context, but wont the backed credentials live in the tenant namespace? For the notebook example , the credential secret  will need be colocated with the with the client that needs it, in its namespace.   Is the proposal here for DCH to automatically copy its the secrets across the cluster to other namespaces that need them?

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,442 @@
...
+## Risks
+
+Rust vs Go can be more challenging for developers that need to ramp up with Rust. However, with AI tools like Claude this can be mitigated significantly with both code and boilerplate generation but also a much faster learning curve for developers. 
+
+## Stakeholder Impacts
+
+| Group                         | Key Contacts     | Date       | Impacted? |
+| ----------------------------- | ---------------- | ---------- | --------- |
+| Data Connect Hub  | Monica Romila, Oronde Tucker, Marius Danciu, Lukasz Cmielowski   | date       | ? |
+
```
lphiri: Please Dashboard,   ODH Platform  and all  current users of existing Data Connections API as stakeholders

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,442 @@
...
+    dataconnect.opendatahub.io/subscription: evals
+rules:
+  # Specify the custom API group defined by your CRD
+  - apiGroups: ["dataconnect.opendatahub.io"]
+    resources: 
+      - api-data-read
+    resourceNames:
+      - {data connection UID 1}
+      - {data connection UID 2}
+    verbs: ["get", "list"]
```
lphiri: Current Data connections support both read and write. 
Is the plan to add write in the future?

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,442 @@
...
+
+##### Hard tenancy
+
+![Fig 2](./images/ADR-0001-img3.png)
+
+
+##### Soft tenancy
+
+![Fig 3](./images/ADR-0001-img4.png)
+
```
lphiri, line 81: If were are going to build this on RHCL/Authorino, we should call out as a consequence that RHCL will be considered a mandatory component to use RHOAI (its currently only needed for llm-d and Maas)

mariusdanciu:
@etirelli @abiazett @lphiri - After several discussion with @danielezonca and also based on this [note](https://redhat-internal.slack.com/archives/C0BHK6U35H9/p1785144613644939) on slack there are some significant changes:

- No Authorino and Limitador - This was mainly a future consideration to align with MaaS however longer term with the Praxis introduction there may be more significant changes such as less reliance on Limitador. The immediate implication is that now we have RBAC rules defined for access different endpoints. Also ConnectionSubscription CR is no longer needed.
- DataConnections as REST resources instead of CRDs - the reasons are discussed in the ADR. The UI considerations also played a significant role (listing, pagination, filtering etc). 
- Migration from the current Connections to DataConnections is on the fly and managed by the DCH Operator.

Please see the updated doc.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: This is similar to evalhub. There is a single DCH service instance hence a single HttpRoute suffice. The connections are identified by id. At REST level the connection id is part of the REST endpoint such as GET `/api/vq/data/connections/{id}`. At Flight gRPC level the connection id is part of the headers. Here is a Python client example:

```Python

conn = dbapi.connect(
    "grpc://127.0.0.1:50051",
    db_kwargs={
        "adbc.flight.sql.rpc.call_header.x-dch-connection-id": "57f315b1-eba0-42a2-bffd-62c03a27a3ee",
        "adbc.flight.sql.rpc.call_header.authorization": "Bearer abc123",
    },
)

info = conn.adbc_get_info()
print("Server Info:")
for key, value in info.items():
    print(f"  {key}: {value}")
print()

cursor = conn.cursor()
cursor.execute("SELECT * FROM prompts")
table = cursor.fetch_arrow_table()
print(table.to_pandas())

```

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Yes as we are adopting soft tenancy for now, we will revisit hard tenancy once this is consolidated at platform level.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: There is one DataConnectService CR per tenant so each tenant gets its own DataConnectService instance in the hard tenancy model. For soft tenancy there is only one DataConnectService in the opendatahub namespace for all tenant. OIDC is not part of MVP so we can separate this into a different CR. Will update the ADR to make this clearer.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Basically `tenant-id = namespace`. So the secrets are managed in the designated namespaces. The DataConnection resources would exist in the same namespace.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Could be but not in our radar yet. We are focusing on ingestion. But yes Arrow Flight can be used for writes as well.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: As per recent discussions we dropped Authorino and Limitador. Kube RBAC proxy for REST and plain SAR from GRPC layer at least until we have Praxis.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: done

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Addressed in the latest changes.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Not applicable since there are no subscriptions anymore.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,458 @@
...
+## Goals
+
+* Define the data ingestion flow for tabular and unstructured data.
+* Define Connections management
+* Define the tenancy model
+* Define access management
+
+
+## Non-Goals
```
andrewballantyne, line 34: Connectivity? "[#wg-openshift-ai-connections-api](https://redhat.enterprise.slack.com/archives/C08T1K4FQVC)" on our slackspace is a room we crafted in RHOAI 3.0 timeframe to talk about how connectivity between a "Connection" and a "Workload" happened.

This is probably a non-goal first swing out -- but it's not called out in either side as I understand it.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,458 @@
...
+
+## How
+
+### Architecture
+
+#### Abbreviations
+
+- DCH = Data Connect Hub
+- DCHO = Data Connect Hub Operator
+- DCHS = Data Connect Service
```
andrewballantyne:
```suggestion
- DCHS = Data Connect Hub Service
```
I assume this was just a typo -- but curious if there is some meaning we are trying to get at with/without "hub"

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,458 @@
...
+We propose to build the DCH solution with multi-tenancy in mind right from its inception. This is a critical aspect that we see growing in the OpenShift ecosystem for other services (MaaS, EvalHub, MLFlow etc.). The core ideas for multi-tenancy are:
+
+##### Hard tenancy
+
+![Fig 2](./images/ADR-0001-img3.png)
+
+
+##### Soft tenancy
+
+![Fig 3](./images/ADR-0001-img4.png)
```
andrewballantyne, line 80: Perhaps this is just me -- but is this a declaration of both Hard & Soft tenancy setups? I'm not sure I'm entirely following how this works -- apologizes if this is a "silly UI architecture question" that is understood by all. I'll keep re-reading this section but it feels you either want to separate tenants explicitly and fully or you don't -- I am not sure why both apply.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,458 @@
...
+              openshift.io/display-name: s3-test
+```
+
+#### REST APIs
+
+This is a non-exhaustive list of endpoints as the set of capabilities is expected to grow.
+
+##### Connections management
+- `GET /api/v1/data/connections` - List connections without sensitive information
+- `POST /api/v1/data/connections` - Create a new connection that points to a pore existent secret.
```
andrewballantyne:
```suggestion
- `POST /api/v1/data/connections` - Create a new connection that points to a prior existing secret.
```
Typo?

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: I'm not sure what you mean.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: We are describing both and the fact that for now we are adopting soft tenancy. Hard tenancy is something that we will/should support in the future.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: yup

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: DCH-Controller will create default roles for writers, readers etc for convenience. For users binding, It is expected that the rolebindings are created manually or via ops tools so that admins grant desired access to the org users.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: The current doc discusses that the current Connections are simply watched and promoted to DataConnection resource that these are just credential secrets. So no explicit migration is needed as this is done automatically. Came would happen with the current ConnectionTypes.

lburgazzoli:
A few small comments:
- the `Tenant DCH Admin Namespace` is a little misleading, in fact it can be any namespace, to my understanding the only restriction is that secrets are created in the same namespace as the `DataConnectionService` CR
- The `Internal CQRS sync` arrow is also probably a leftover, it would probably be good to describe the high level interaction between the Data Connection Operator and the Data Connection Services

mariusdanciu:
> A few small comments:
> 
> * the `Tenant DCH Admin Namespace` is a little misleading, in fact it can be any namespace, to my understanding the only restriction is that secrets are created in the same namespace as the `DataConnectionService` CR
> * The `Internal CQRS sync` arrow is also probably a leftover, it would probably be good to describe the high level interaction between the Data Connection Operator and the Data Connection Services

@lburgazzoli thank you. I updated the diagram's labels:

1. I used the term " tenant namespace". it is basically the namespace where secrets are managed.
2. CQRS is indeed not needed anymore since DataConnection and DataConnectionTypes are REST resources not CRDs anymore. That arrow is still needed as the controller needs to inform the service when hydration from existenf Connection secrets or Connation Type configmaps need to happen. Also for communicating when an actual secret was removed so we can mark the DataConnection as not-ready.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
andrewballantyne: In our current world -- we have 3 layers.

1. A definition / type => Connection Types
2. An instance (of a type) => Connections
3. A usage of an instance / "connectivity" => Connection-API (annotations on specific resources)

no3 is not used by a couple places -- MR & DPSA -- but that's because they don't actually use Connections for varying reasons.

no3 is also what I was referring to -- how to connect to a Workbench or a Model Serving instance (aka workloads), is an important point here. I think you should add it to the non-goals based on our chats in the meeting yesterday. @mariusdanciu 

If this is still a weird ask -- lets quickly meet and I can explain more.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,466 @@
...
+  name: data-connect-hub
+  namespace: opendatahub
+spec:
+  description: Connection to the test data S3 bucket
+  restApiReplicas: 3 # defaults to 2
+  flightApiReplicas: 3 # defaults to 3
+
+status:
+  phase: Ready
+  externalHostName: example.myorg.com
```
danielezonca: Is this the URL to reach the service?
It is common to use the convention of `status.address.url` for this information

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,466 @@
...
+
+- `GET /api/v1/data/connection_types` - List connection types 
+- `POST /api/v1/data/connections_types` - Create a new connection type.
+- `PATCH /api/v1/data/connection_types/{id}` - Update a connection type 
+- `GET /api/v1/data/connection_types/{id}` - Get the details of a specific connection type
+- `DELETE /api/v1/data/connection_types/{id}` - Delete a connection type
+
+  Currently Connection Types in RHOAI are represented as configmaps and they are very UI driven. We aim here for:
+  1. Have a type safe API for managing connection types REST resources
+  2. On the fly promotion of current configmaps to actual connection types resources. Once synced, these connections types are visible to all tenants (cluster scoped).
```
danielezonca, line 201: This is outside the scope of the ADR but I just want to highlight about the UI given that the new connection will replace the existing once GA:
- if you have new connections active there is a "dev preview" label and a note in the page like "connections will be merged and replace existing connections"
- same if you are in "old" connection and new connections are active you will see some note "try new connections"

@chrjones-rh @andrewballantyne 
FYI

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,466 @@
...
+
+```
+  DataConnection + credentials secret = Exported Secret
+```
+
+The admin can control the RBAC rules that are applied for the new secret and who can use it. At this point, DCHO will not automatically create and manage RBAC for this secret. It is up to the admin to manage RBAC for an Exported-Secret. The DataConnection resource needs to be configured in order to allow the auto creation of the new secret. Thus, the resource would need to contain the `exportAsSecret` flag:
```
danielezonca, line 337: Is the exported secret related to the tenancy model described above?
Because secrets are namespaces scoped while I'm not sure we are making the same assumption for the tenancy

danielezonca (approved):
overall lgtm, just some minor comments.
once resolved I'm fine to get this merged

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: Yes the exported secret is always in the same namespace with the DataConnection tenant. We are not crossing namespaces.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: > if you have new connections active there is a "dev preview" label and a note in the page like "connections will be merged and replace existing connections"

>same if you are in "old" connection and new connections are active you will see some note "try new connections"

I agree with both statements that the UI should do this. And yes this is outside the scope of this ADR. @chrjones-rh FYI.

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: The Gateway CR uses :
```
Status:
  Addresses:
    Type:   Hostname
    Value:  data-science-gateway-data-science-gateway-class.openshift-ingress.svc.cluster.local
```
 MaasModelRef uses `status.httproutehostnames`.

I like the Gateway approach better. I will update the doc.

architecture-decision-records/data-connect-hub/images/ADR-0001-img3.png
lphiri, line 1: @mariusdanciu  I think you meant "DataConnectHub Service instance tenat 2"  (not tenant 1) for the bottom one?

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
```diff
@@ -0,0 +1,469 @@
...
+  labels:
+    dataconnecthub.opendatahub.io/connection_id=1fac3d4-e5f6-7a8b-9cff-1e2f3a4b5caa
+    dataconnecthub.opendatahub.io/connection_name=my_eval
+
+type: Opaque
+Data:
+  {KV properties here}
+```
+
+**Note** - The exported secret will always live in the same namespace with the credential secret. Thus the secret information never crosses the namespace boundaries.
```
lphiri, line 371: @mariusdanciu how do we map a given discovered secret to a particular `tenant-id` in the metadata store?
I was doing a mental walkthrough of how we would use the `tenant-id` header to filter for example:
GET /data/connections 
and how that is related to the rbac roles described above.

architecture-decision-records/data-connect-hub/images/ADR-0001-img3.png
mariusdanciu: Typo yes. Good eyes !

architecture-decision-records/data-connect-hub/ODH-ADR-0001-data-connect-hub.md
mariusdanciu: In DB the secret name is referenced by the DataConnection REST object as `"secret_ref": "my_secret"`. The DataConnection REST object is stored in DB with the tenant_id. so the DataConnection REST object tenant_id matches with the namespace of the secret. So

```
GET /data/connections
X-Tenant-Id: team_a

```

Will return the data-connections from team_a namespace. The RBAC will be checked agains the RoleBindings in team_a namespace at kube RBAC proxy level.

lphiri (approved):

architecture-decision-records/data-connect-hub/images/ADR-0001-img3.png
mariusdanciu: Corrected.

PR #150, opened 2026-07-23
ODH-ADR-DR-0001: Data Registry for RHOAI

abiazett:
## Summary

- Adds a new `data-registry` folder under `architecture-decision-records/`
- Introduces **ODH-ADR-DR-0001**: Data Catalog and Data Registry for RHOAI

### Decision

Adopt the **Iceberg REST Catalog API** as the standard catalog interface for RHOAI, backed by a Feast registry translation layer in Phase 1. Key architectural choices:

- **Reuse & Extend** — dedicated catalog server pod via `spec.catalog.enabled` on existing FeatureStore CRD (no new CRDs or operators)
- **SAR RBAC** — platform-native access control via SubjectAccessReview (proven RHAI pattern via kube-rbac-proxy)
- **Data Hub UI** — new dashboard module for datasets, collections, and volumes
- **Phased evolution** — translation layer now, native Iceberg backends (UC OSS, Polaris) evaluated in Phase 2+

### Status

Under Review

### Companion documents

Internal analysis docs (design options, concept mapping, interoperability analysis, RBAC design) are referenced in the ADR and available in the [data-strategy GitLab group](https://gitlab.cee.redhat.com/data-strategy/data-arch-design-options).

## Summary by CodeRabbit

* **Documentation**
  * Added an ADR describing the Phase 1 Data Registry MVP for RHOAI, including an Iceberg REST catalog–compatible subset plus Data Registry extensions (volumes and search).
  * Documented Data Hub UI integration via a dashboard BFF that forwards OAuth bearer tokens to the Data Registry API.
  * Defined Kubernetes-native access control using SAR via kube-rbac-proxy, including aggregated view/edit/admin semantics and catalog-only sharing to limit access scope.
  * Specified operational guidance: network policies, timeouts/pagination, observability signals, rollback via feature-flagged deployment, and a future migration path to a native Iceberg backend.

PR #153, opened 2026-08-05
ADR-DR-0002: Data Registry and Data Connect Hub Integration

abiazett:
## Summary

- Defines the integration contract between the Data Registry (ADR-DR-0001) and the Data Connect Hub (DCH) for the 6.3 release
- Introduces the `connection_ref` model with a type discriminator (`dch` / `rhai`) and connection identifier (UUID / secret name)
- Documents three consumption scenarios: Iceberg table with S3 (mounted secret), PostgreSQL via DCH ingestion (Arrow Flight), and local PVC volume
- Establishes the shared SSAR authorization model across both API groups (`dataregistry-view/edit/admin` + `dataconnecthub.opendatahub.io` resources)

## Related

- ADR-DR-0001: Data Registry for RHOAI (#150)
- DCH ADR (#149)
- ADR-DR-0003: Future Integration Scenarios (separate PR)

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record describing optional integration between the Data Registry and Data Connect Hub.
  * Documented supported data access scenarios, including asset discovery, Iceberg/S3 access, PostgreSQL ingestion, and local storage access.
  * Clarified connection references, authorization, credential isolation, migration considerations, risks, security requirements, and implementation guidance.
  * Included guidance for TLS, credential retrieval, and direct versus hub-mediated connections.

### DIFF
--- a/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
+++ b/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
@@ -85,6 +85,10 @@ By default, the DCH DataConnection response does not include credential values �
 
 
 
+### Client Integration for 3.6
+
+For 3.6, the client (notebook, pipeline, agent) performs the multi-step flow: discover via Data Registry API, resolve connection via DCH API, access data. A platform SDK that orchestrates this flow into a single call is a future enhancement (see [Alternative 2](#alternative-2-credential-vending-via-data-registry-loadtable) and [ADR-DR-0003](https://github.com/opendatahub-io/architecture-decision-records/pull/154)).
+
 ### Scenario 0: Discovering a Data Asset
 
 A data scientist working in a workbench pod wants to find claims-related data. They search the Data Registry using the Data Registry API, find the asset they need, and retrieve its full metadata — including schema, connection reference, and properties.
@@ -197,6 +201,8 @@ Local storage is managed by OpenShift (PersistentVolume / PersistentVolumeClaim)
 | DCH DataConnection exists but is not ready | UI shows connection status. Asset is browsable in the Data Registry, but accessing data may fail. UI warns about connection state. |
 | Asset registered without `connection_ref` | Valid for local PVC volumes and assets where the user manages access independently. |
 | RHAI Connection after DCH auto-migration | Both `type: rhai` and `type: dch` references remain valid. The underlying Kubernetes secret is preserved by DCH during migration. |
+
+**Connection status visibility:** The Data Registry stores `connection_ref` as metadata — it does not track or validate the connection's runtime status. When an asset has a `connection_ref`, the consuming client (notebook, pipeline, agent, Data Hub UI) must call the DCH API (`GET /api/v1/data/connections/{id}`) to determine whether the connection exists, is ready, or is accessible to the current user. The Data Registry returns the asset regardless of connection state; it is the responsibility of the consuming client to check connection status and surface appropriate guidance — such as "connection not found", "connection not ready", or "insufficient permissions" — before the user attempts to access the data.
 
 ## Alternatives
 
--- a/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
+++ b/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
@@ -77,7 +77,7 @@ DCH auto-migrates existing RHAI Connections by watching labeled Kubernetes secre
 
 For Scenario 1, users need direct access to connection credentials (e.g., S3 keys) without relying on DCH ingestion APIs. Rather than requiring users to mount the connection secret onto their pod — which can cause conflicts when multiple connections use the same environment variable names (e.g., `AWS_ACCESS_KEY_ID`) — DCH supports retrieving credentials programmatically:
 
-```
+```http
 GET /api/v1/data/connections/{id}?with_secret_info=true
 ```
 
@@ -205,8 +205,9 @@ Local storage is managed by OpenShift (PersistentVolume / PersistentVolumeClaim)
 | DCH DataConnection exists but is not ready | UI shows connection status. Asset is browsable in the Data Registry, but accessing data may fail. UI warns about connection state. |
 | Asset registered without `connection_ref` | Valid for local PVC volumes and assets where the user manages access independently. |
 | RHAI Connection after DCH auto-migration | Both `type: rhai` and `type: dch` references remain valid. The underlying Kubernetes secret is preserved by DCH during migration. |
+| `connection_ref` type does not match the asset format | The consuming client must validate that the connection type is compatible with the asset format before resolving credentials — e.g., reject PostgreSQL connection credentials for an Iceberg table, or S3 credentials for a PostgreSQL ingestion. Mismatched references should be surfaced as a validation error, not silently resolved. |
 
-**Connection status visibility:** The Data Registry stores `connection_ref` as metadata — it does not track or validate the connection's runtime status. When an asset has a `connection_ref`, the consuming client (notebook, pipeline, agent, Data Hub UI) must call the DCH API (`GET /api/v1/data/connections/{id}`) to determine whether the connection exists, is ready, or is accessible to the current user. The Data Registry returns the asset regardless of connection state; it is the responsibility of the consuming client to check connection status and surface appropriate guidance — such as "connection not found", "connection not ready", or "insufficient permissions" — before the user attempts to access the data.
+**Connection status visibility:** The Data Registry stores `connection_ref` as metadata — it does not track or validate the connection's runtime status. When an asset has a `connection_ref`, the consuming client (notebook, pipeline, agent, Data Hub UI) resolves connection status through the path appropriate for the connection type: for `type: dch`, call the DCH API (`GET /api/v1/data/connections/{id}`) to determine whether the connection exists, is ready, or is accessible to the current user; for `type: rhai`, check the referenced Kubernetes secret directly in the asset's namespace. The Data Registry returns the asset regardless of connection state; it is the responsibility of the consuming client to check connection status and surface appropriate guidance — such as "connection not found", "connection not ready", or "insufficient permissions" — before the user attempts to access the data.
 
 ## Alternatives
 
@@ -241,7 +242,7 @@ The Data Registry defines three ClusterRoles that grant access to all registry r
 
 | ClusterRole | Access Level |
 |---|---|
-| `dataregistry-view` | Read-only access to all Data Registry resources (tables, volumes, collections, search) |
+| `dataregistry-view` | Read-only access to all Data Registry resources (namespaces, tables, volumes) |
 | `dataregistry-edit` | Read and write access to Data Registry resources |
 | `dataregistry-admin` | Full access including administrative operations |
 
@@ -281,7 +282,17 @@ The table above covers the consumption side — discovering and using data. Regi
 
 DCH does not introduce new K8s Secret permissions — credentials still live in standard Kubernetes secrets, and mounting or reading them requires the same `secrets: get` permission as any other secret. Creating a DataConnection via the DCH API requires `data-connection: create`; if the underlying secret does not yet exist, `secrets: create` is also needed.
 
-Both components use ClusterRole aggregation labels to auto-inject permissions into standard OpenShift roles (`view`, `edit`, `admin`). Users with existing namespace roles automatically receive the corresponding Data Registry and DCH permissions without additional RBAC configuration. The specific aggregation label selectors for each component are TBD — to be defined in the respective Data Registry and DCH ADRs.
+Both components use ClusterRole aggregation labels to auto-inject permissions into standard OpenShift roles (`view`, `edit`, `admin`). Users with existing namespace roles automatically receive the corresponding Data Registry and DCH permissions without additional RBAC configuration.
+
+**Aggregated roles summary:**
+
+| OpenShift Role | Data Registry ClusterRole | Data Registry Verbs | DCH Permissions |
+|---|---|---|---|
+| `view` | `dataregistry-view` | `get`, `list` on namespaces, tables, volumes | `data-connection: read`, `data-store: get` |
+| `edit` | `dataregistry-edit` | `get`, `list`, `create`, `update`, `delete` on namespaces, tables, volumes | `data-connection: create, read, patch, delete`, `data-store: get` |
+| `admin` | `dataregistry-admin` | Same as edit (future: grant/revoke capabilities) | Full DCH access |
+
+For most deployments, this means zero additional RBAC configuration — users who can already work in a namespace can automatically use the Data Registry and DCH in that namespace.
 
 ### Credential Isolation
 
--- a/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
+++ b/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
@@ -17,7 +17,7 @@ This ADR defines the integration contract between the **Data Registry** ([ADR-DR
 
 The integration is based on a `connection_ref` field that each Data Registry asset (table or volume) can optionally carry, pointing to a DCH DataConnection or an RHAI Connection. The `connection_ref` includes a type discriminator (`dch` or `rhai`) and the connection identifier (UUID for DCH, secret name for RHAI). In both cases, `connection_ref` references a *connection object* — not raw credentials. The actual secret is resolved through the respective connection API. This ADR defines the connection reference contract and three consumption scenarios for the 6.3 release.
 
-Future integration scenarios (automated schema discovery and cross-component lineage via OpenLineage) are defined in [ADR-DR-0003](https://github.com/opendatahub-io/architecture-decision-records/pull/TBD).
+Future integration scenarios (automated schema discovery and cross-component lineage via OpenLineage) are defined in [ADR-DR-0003](https://github.com/opendatahub-io/architecture-decision-records/pull/154).
 
 ## Why
 
@@ -39,8 +39,8 @@ The Data Registry and DCH are being developed in parallel. Defining the integrat
 
 * **Changing the internal architecture of either component.** The Data Registry and DCH internals are defined by their own ADRs. This ADR covers only the integration boundary.
 * **New APIs for 6.3.** No new Data Registry or DCH APIs are introduced for 6.3. The `connection_ref` attribute on tables and volumes is a metadata field, not a new API surface.
-* **Schema synchronization in 6.3.** Automated schema discovery from data sources is a future scenario documented in [ADR-DR-0003](https://github.com/opendatahub-io/architecture-decision-records/pull/TBD).
-* **Lineage tracking across the integration boundary in 6.3.** Cross-component lineage via OpenLineage is out of scope for 6.3, documented in [ADR-DR-0003](https://github.com/opendatahub-io/architecture-decision-records/pull/TBD).
+* **Schema synchronization in 6.3.** Automated schema discovery from data sources is a future scenario documented in [ADR-DR-0003](https://github.com/opendatahub-io/architecture-decision-records/pull/154).
+* **Lineage tracking across the integration boundary in 6.3.** Cross-component lineage via OpenLineage is out of scope for 6.3, documented in [ADR-DR-0003](https://github.com/opendatahub-io/architecture-decision-records/pull/154).
 
 ## How
 
@@ -248,7 +248,7 @@ Both components use ClusterRole aggregation labels to auto-inject permissions in
 ## References
 
 * [ADR-DR-0001: Data Registry for RHOAI (PR #150)](https://github.com/opendatahub-io/architecture-decision-records/pull/150)
-* [ADR-DR-0003: Future Integration Scenarios (PR #TBD)](https://github.com/opendatahub-io/architecture-decision-records/pull/TBD)
+* [ADR-DR-0003: Future Integration Scenarios (PR #TBD)](https://github.com/opendatahub-io/architecture-decision-records/pull/154)
 * [Data Connect Hub ADR (PR #149)](https://github.com/opendatahub-io/architecture-decision-records/pull/149)
 * [Iceberg REST Catalog Spec](https://iceberg.apache.org/rest-catalog-spec/#rest-catalog-protocol)
 * [opendatahub-io/kube-rbac-proxy](https://github.com/opendatahub-io/kube-rbac-proxy) — SSAR authorization sidecar
--- a/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
+++ b/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
@@ -65,7 +65,7 @@ In the table below we list the different Data Registry Asset Types and formats, 
 | Table | iceberg | S3 credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (PyIceberg, Spark, Trino) |
 | Table | postgresql | DCH DataConnection | `{type: dch, id: uuid}` | DCH SDK |
 | Table | oci (OCI Artifact registry) | Credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (user libraries) |
-| Volume | s3 | S3 credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (user libraries) |
+| Volume | s3 | S3 credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | DCH SDK (jsonl, binary via Flight) or Direct (user libraries) |
 | Volume | uri/pvc | Credentials via DCH API or RHAI secret, or OpenShift-managed (local PVC) | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` or None (local) | Direct (user libraries / filesystem) |
 | Volume | hugging face (custom uri) | Credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (user libraries) |
 
@@ -157,7 +157,7 @@ sequenceDiagram
     DR-->>User: table metadata<br/>format: postgresql<br/>connection_ref: {type: dch,<br/>id: 550e8400-...}<br/>schema: [claim_id, status, amount]
 
     User->>DCH: dch.ingest(connection_id="550e8400-...",<br/>query="SELECT * FROM claims")
-    Note over User,DCH: Headers: Authorization, x-tenant-id,<br/>x-dch-connection-id
+    Note over User,DCH: Headers: Authorization, x-tenant-id,<br/>x-data-connection-id
 
     DCH->>PG: Connect using stored credentials<br/>Execute query
     PG-->>DCH: Result rows
@@ -167,7 +167,7 @@ sequenceDiagram
 The user never sees database credentials. DCH handles driver dependencies, connection pooling, and credential management. The Data Registry provides the `connection_ref` that links the table asset to the right DCH DataConnection.
 
 This pattern is essential for:
-- Non-S3 data sources (e.g., PostgreSQL) where credential mounting alone is insufficient for data ingestion
+- Data sources where users benefit from DCH handling the connection and data transfer (e.g., PostgreSQL, S3 with jsonl/binary formats)
 - Environments where credential exposure to user pods must be minimized
 - Clients that do not want to install and maintain database-specific driver libraries
 
@@ -221,6 +221,8 @@ Instead of retrieving credentials programmatically via `?with_secret_info=true`,
 Iceberg engines (PyIceberg, Spark, Trino) expect `loadTable` to return credentials in the response `config` field — this is standard Iceberg REST Catalog credential vending (how Unity Catalog and Polaris work). If DCH had a credential resolution API, the Data Registry `loadTable` could call it internally and return credentials in `config`. Engines would get everything in one call — no separate DCH API call, no mounting secrets, no workbench restart.
 
 This approach would improve Scenario 1 by eliminating the extra step of calling the DCH API with `?with_secret_info=true` — credentials would be vended as part of the standard Iceberg REST Catalog flow. However, it requires a new server-to-server integration between the Data Registry and DCH (the Registry would need to call DCH's credential resolution API during `loadTable`), which introduces a runtime dependency and is out of scope for the current release. This is tracked as a future integration scenario in [ADR-DR-0003](https://github.com/opendatahub-io/architecture-decision-records/pull/154).
+
+**Scoping note:** DCH returns credentials from Kubernetes secrets only (given proper RBAC permissions). There is no plan in DCH to integrate with cloud-native credential vending services (AWS STS, IBM Cloud IAM, GCP, Azure) or Vault. A future `loadTable` credential vending integration would be scoped to Kubernetes secret-based credentials resolved through DCH.
 
 ## Security and Privacy Considerations
 
--- a/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
+++ b/architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
@@ -23,7 +23,7 @@ Future integration scenarios (automated schema discovery and cross-component lin
 
 The Data Registry and DCH are being developed in parallel. Defining the integration contract now — before either component ships — ensures users get a coherent "find to use" experience from the first release, rather than requiring a retrofitted integration later.
 
-- **Unified "find to use" experience.** The product requirement is that discovering an asset and accessing its data should feel like one action, not two disconnected systems.
+- **Unified "find to use" experience.** The product requirement is that discovering an asset and accessing its data should work as an unified and consistent flow, although underlying systems may be different.
 - **Connection flexibility.** DCH connections must work like RHAI Connection secrets at the basic level — mountable on pods for direct data access via standard libraries. For clients that need mediated access to data sources (e.g., PostgreSQL via Arrow Flight), DCH provides ingestion APIs that handle driver complexity and credential isolation.
 - **Consistent authorization.** Both the Data Registry and DCH use SubjectAccessReview (SAR) via kube-rbac-proxy with Kubernetes-native RBAC. The auth model is already aligned, but the integration contract — which permissions are needed for the end-to-end flow — must be explicit.
 - **Local storage independence.** Not all data requires external connections. Volumes backed by PVC storage use OpenShift-managed credentials with no DCH involvement. The Registry must work both with and without DCH.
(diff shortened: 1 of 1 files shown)


architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+
+### Connection Reference Model
+
+The Data Registry `connection_ref` field on tables and volumes stores a reference to a connection object — either a DCH DataConnection (identified by UUID) or an RHAI Connection (identified by secret name). The field is optional — assets with locally mounted PVC volumes do not require a connection reference.
+
+`connection_ref` always points to a connection *object*, never to raw credentials. It includes two subfields:
+
+| Subfield | Description | Example |
+|---|---|---|
+| `type` | Discriminator: `dch` or `rhai` | `dch` |
```
mariusdanciu, line 54: I wonder why support both old and new. The idea is that current Connections would be replaced by DCH. So is this a temporary solution because DataRegistry is not DP/TP and DCH is ? Aren't both DataRegistry and DCH both new features ?

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+- **`rhai`:** The `id` is the name of a labeled Kubernetes secret (`opendatahub.io/dashboard=true`, `opendatahub.io/managed=true`). The secret is resolved directly via the Kubernetes API.
+
+In the table below we list the different Data Registry Asset Types and formats, and which connection pattern they support as well as whether DCH will support their direct Ingestion (in 3.6).
+
+| Asset Type | Format | Connection Pattern | connection_ref | Ingestion |
+|---|---|---|---|---|
+| Table | iceberg | S3 credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (PyIceberg, Spark, Trino) |
+| Table | postgresql | DCH DataConnection | `{type: dch, id: uuid}` | DCH SDK |
+| Table | oci (OCI Artifact registry) | Credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (user libraries) |
+| Volume | s3 | S3 credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (user libraries) |
```
mariusdanciu: DCH will support S3 ingestion via flight for some file formats (i.e. jsonl) also for direct S3 file download as binary.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+
+**Retrieving credentials from a DCH DataConnection:**
+
+For Scenario 1, users need direct access to connection credentials (e.g., S3 keys) without relying on DCH ingestion APIs. Rather than requiring users to mount the connection secret onto their pod — which can cause conflicts when multiple connections use the same environment variable names (e.g., `AWS_ACCESS_KEY_ID`) — DCH supports retrieving credentials programmatically:
+
+```
+GET /api/v1/data/connections/{id}?with_secret_info=true
+```
+
+By default, the DCH DataConnection response does not include credential values — the admin/secret section is omitted for regular users. If the caller has RBAC permissions to read secrets (`secrets: get` in the namespace), the `?with_secret_info=true` query parameter instructs DCH to unpack the secret keys and include them in the response. This allows users to retrieve credentials on demand without pod-level secret mounting, avoiding environment variable collisions across multiple connections.
```
mariusdanciu, line 84: All this looks correct to me with one additional detail. In order to be able to retrieve the actual credentials, the identity must have the RBAC permission to `get` `secrets` resources. This API behavior helps the client (or UI) to not do a separate kube api call to read the secret details (if allowed ot).

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+    participant User as Data Scientist<br/>(Workbench Pod)
+    participant DR as Data Registry API
+    participant DCH as DCH Service<br/>(Arrow Flight)
+    participant PG as PostgreSQL
+
+    User->>DR: GET /v1/{p}/namespaces/{c}/generic-tables/auto-claims-db
+    DR-->>User: table metadata<br/>format: postgresql<br/>connection_ref: {type: dch,<br/>id: 550e8400-...}<br/>schema: [claim_id, status, amount]
+
+    User->>DCH: dch.ingest(connection_id="550e8400-...",<br/>query="SELECT * FROM claims")
+    Note over User,DCH: Headers: Authorization, x-tenant-id,<br/>x-dch-connection-id
```
mariusdanciu: Although this is correct I am thinking to rename `x-dch-connection-id` header to `x-data-connection-id` header. And update this in the DCH ADR as well.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+    User->>DR: GET /v1/{p}/namespaces/{c}/volumes/claim-documents
+    DR-->>User: volume metadata<br/>connection_ref: {type: dch,<br/>id: 550e8400-...}
+
+    Note over User: User resolves PVC location<br/>from connection_ref metadata.<br/>User must ensure PVC is<br/>mounted on their pod<br/>(via workbench storage config<br/>or pod volume spec)
+
+    User->>PVC: Read files from /mnt/data/documents/
+    PVC-->>User: File contents
+```
+
+Local storage is managed by OpenShift (PersistentVolume / PersistentVolumeClaim). The `connection_ref` points to the connection describing the PVC, but no credential retrieval via DCH is needed — the user is responsible for ensuring the PVC is mounted on their pod, via workbench storage configuration in the RHOAI Dashboard or by specifying the volume in their pod spec.
```
mariusdanciu, line 195: Does the user need to use PVC in order to mount DataConnection secrets ? 

I think a 4th scenario for data ingestion may not imply data-registry at all. The user simply mounts the DataConnection secret in their POD (i.e. notebook) and read the data. This is what happens today as well. This is addressed in Alternative 1 below but this is not something we support or not. This is what users can to do already.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+    User->>DCH: GET /api/v1/data/connections/550e8400-...?with_secret_info=true
+    DCH-->>User: DataConnection metadata +<br/>credentials (keys, host,<br/>username, password, etc.)
+
+    Note over User: Credentials retrieved<br/>programmatically — no pod<br/>secret mounting needed
+
+    User->>DS: Access data source directly<br/>using retrieved credentials<br/>(PyIceberg, Spark, Trino, etc.)
+    DS-->>User: Data
+```
+
+The user resolves the `connection_ref` by calling the DCH API with `?with_secret_info=true`, which returns the credential values inline (requires `secrets: get` RBAC permission). This avoids the need to mount connection secrets on the pod. DCH's role is managing the connection lifecycle (creation, credential rotation, auto-migration from RHAI Connections) and serving credentials on demand. The Data Registry's role is providing the table schema and metadata.
```
mariusdanciu, line 143: This implies that the user use DCH SDK or REST API to retrieve the credentials. While this resolves some challenges with envvar mouting for instance, some clients may not be interested in using other APIs. We cannot stop them from mounting the secrets directly as this is entirely up to them

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+
+## Alternatives
+
+### Alternative 1: Mounting Connection Secrets on Pods
+
+Instead of retrieving credentials programmatically via `?with_secret_info=true`, users mount connection secrets directly on their pods as environment variables — either by attaching an RHAI Connection to a workbench via the RHAI Dashboard, or by mounting a DCH ExportedSecret via pod volume spec.
+
+**Why not chosen as the primary approach:**
+
+- **Environment variable collisions.** Mounting connection secrets as environment variables on the pod causes conflicts when multiple connections use the same variable names (e.g., `AWS_ACCESS_KEY_ID`). RHAI connections currently enforce a restriction that only one connection of the same type (e.g., S3) can be attached to a workbench, specifically because of these environment variable name collisions. This makes it impractical for users who need to access multiple S3-compatible data sources from the same workbench.
```
mariusdanciu: Yes this is a challenge but can be addressed at kube level either by mounting the secrets as files in different paths or mounting them as env_var but use prefixes in the kube descriptor.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+- **Workbench restart required.** Currently, a workbench must be restarted for new secrets to be attached to the pod. This means a user discovering a new data asset in the Data Registry would need to stop their workbench, attach the connection, restart, and resume their work — making the "find to use" flow disruptive and impractical.
```
mariusdanciu: Restarts are not required id secrets are mounted as files. Only if they are mounted as env-vars. But this is a workbench decision of course.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+- **ExportedSecret complexity.** The DCH ExportedSecret approach (where the DCH operator creates a combined secret from DataConnection metadata and credentials) adds operational complexity — an additional operator watch loop, a new secret resource to manage, and the same pod-mounting limitations above.
```
mariusdanciu: I'm not quite following where the complexity it. It is not the dch-controller that creates the exported secret, it is the dch-service api layer (but this is not very relevant). If a DataConnection is marked as "exportAsSecret" we automatically create the exported secret hence it is ready to use. The user tells us the secret name so the user knows directly where to find it. it is really very simple. 

Regarding pod-mounting limitations, yes of course since it it a secret mounting happens via kube machinery. But there are ways to address the challenged like where a POD needs to mount more than one secret. Generally I avoid using secrets env_var because this raises security concerns, pod restarts etc. But it is ultimately the user's decision how to consume secrets.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
mariusdanciu: Note that this option will only work for users that have the RBAC permission to read secrets information.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+
+    DCH->>PG: Connect using stored credentials<br/>Execute query
+    PG-->>DCH: Result rows
+    DCH-->>User: Arrow RecordBatch
+```
+
+The user never sees database credentials. DCH handles driver dependencies, connection pooling, and credential management. The Data Registry provides the `connection_ref` that links the table asset to the right DCH DataConnection.
+
+This pattern is essential for:
+- Non-S3 data sources (e.g., PostgreSQL) where credential mounting alone is insufficient for data ingestion
```
mariusdanciu: Why for non S3 data source. We are planning to support S3 connector with certain file types formats like jsonl. Other file formats like parquet, csv etc may come in a later phase.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,316 @@
...
+- **Workbench restart required.** Currently, a workbench must be restarted for new secrets to be attached to the pod. This means a user discovering a new data asset in the Data Registry would need to stop their workbench, attach the connection, restart, and resume their work — making the "find to use" flow disruptive and impractical.
+- **ExportedSecret complexity.** The DCH ExportedSecret approach (where the DCH operator creates a combined secret from DataConnection metadata and credentials) adds operational complexity — an additional operator watch loop, a new secret resource to manage, and the same pod-mounting limitations above.
+
+### Alternative 2: Credential Vending via Data Registry loadTable
+
+Iceberg engines (PyIceberg, Spark, Trino) expect `loadTable` to return credentials in the response `config` field — this is standard Iceberg REST Catalog credential vending (how Unity Catalog and Polaris work). If DCH had a credential resolution API, the Data Registry `loadTable` could call it internally and return credentials in `config`. Engines would get everything in one call — no separate DCH API call, no mounting secrets, no workbench restart.
```
mariusdanciu: DCH will only return credentials from secrets (given propose RBAC permissions). There is no plan in DCH to support credentials vending and integrate with AWS STS, IBM Cloud IAM, GCP, Azure etc. or even with Vault vending support. This is not in our radar at this point.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Good question. Two reasons for supporting both:

a) **Data Registry is TP in 3.6 and needs to work without DCH.** Since Data Registry ships independently, it must support current RHAI Connections out of the box — users shouldn't need to wait for DCH to register data assets.

b) **Not all RHAI connection types may be supported by DCH right away.** A customer may still want to register a Data Asset on the Data Registry pointing to an RHAI Connection for a connection type that DCH doesn't yet cover.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Good point — will update the Connection Reference Model table to reflect that DCH supports S3 ingestion via Arrow Flight for certain file formats (jsonl, binary download). The table will show both options: DCH SDK (via Flight) and Direct (user libraries), since users may still choose to access S3 directly with their own libraries.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Agreed — this is already covered in the ADR. Line 84 explicitly states the `secrets: get` RBAC requirement, and the permissions table (line 273) includes it as well. The API convenience of avoiding a separate kube API call is a good way to frame it.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Makes sense — will update the Scenario 2 diagram to use `x-data-connection-id`.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: You're right — DCH supports S3 ingestion for certain file formats (jsonl initially, with parquet/csv in later phases). Will reword this to remove the "Non-S3" qualifier. The point is that delegated ingestion via DCH is essential for data sources where the user benefits from DCH handling the connection and data transfer, regardless of the underlying storage type.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Noted — will add an explicit scoping note to Alternative 2 clarifying that DCH returns credentials from Kubernetes secrets only (given proper RBAC permissions), and does not integrate with cloud-native credential vending services (AWS STS, IBM Cloud IAM, GCP, Azure) or Vault. This correctly scopes what a future `loadTable` integration could deliver.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
MStokluska: Hey @mariusdanciu 
Can you clarify on how to achieve this without restart? From my understanding, to mount a secret as file you still need to mount a volume. After that, no further restarts are required but in the flow that we are proposing: discover -> use, that restart will still be required. Am I missing something?

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: I assume by Data Connection you mean current RHAI Connection? If so, then yes, I will add comment on scenario 3 to clarify that a) not just PVC can use RHAI Connections directly and b) Data Registry does not need to be involved in the scenario, the user can continue to use RHAI connections and mount it as they do it today.

abiazett: Good point. This scenario is not limited to PVC — any RHAI Connection type can follow the same pattern where the user mounts the connection secret directly on their pod. This is the default behavior today and continues to work regardless of whether the Data Registry is involved. Will add a clarifying paragraph to Scenario 3 making both points explicit.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
MStokluska: I wonder if we should really even mention the workaround because:

1. It raises the bar when it comes to required knowledge when using the data registry (k8s knowledge required to edit Notebook CRs manually).
2. Manual changes to Notebook CRs do not sync with the Dashboard UI. This creates a source-of-truth problem - the UI shows one set of mounted connections while the actual pod has additional mounts the UI doesn't know about.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Agreed — direct secret mounting is absolutely a valid and supported path. The `?with_secret_info=true` flow is an additional option, not a replacement. I've added a note at the end of Scenario 1 clarifying that clients may continue to use the current approach of mounting connection secrets as environment variables or files, and pointing to Alternative 1 which documents the trade-offs of that approach.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: @mariusdanciu does DCH plan to improve on the secret mounting  (and attachment to a workbench) capabilities by supporting the use of prefixes or secret mounting as files, or is that expected to be handled by the user?

abiazett: Good point — I've updated Alternative 1 to acknowledge that file mounts and env var prefixes are valid workarounds for the collision issue. Alternative 1 is now framed as the current RHAI Connections capability that clients can continue to use, with the known restrictions noted rather than presented as reasons for rejection.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Agreed — I've updated the restart bullet in Alternative 1 to clarify: a restart is needed to attach a *new* connection (env vars or files), but for secrets already mounted as files, the kubelet updates contents automatically on credential rotation without requiring a restart.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Thanks for the clarification — I've rewritten the ExportedSecret section in Alternative 1. It now correctly states that when a DataConnection is marked as `exportAsSecret`, the DCH service API layer automatically creates the exported secret and it is ready to use. The user specifies the secret name, so they know directly where to find it.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: Agreed — this is already noted in the ADR: the `?with_secret_info=true` flow requires `secrets: get` RBAC permission, as stated in the Scenario 1 description and the permissions table.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
abiazett: I updated the whole subsection to clarify the intent and take into account your points

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,301 @@
...
+
+## What
+
+This ADR defines the integration contract between the **Data Registry** ([ADR-DR-0001](https://github.com/opendatahub-io/architecture-decision-records/pull/150)) and the **Data Connect Hub** (DCH, [PR #149](https://github.com/opendatahub-io/architecture-decision-records/pull/149)). The Data Registry tracks *what* data exists — tables, volumes, metadata, schema. DCH manages *how* to connect to and ingest data from external sources.
+
+The integration is based on a `connection_ref` field that each Data Registry asset (table or volume) can optionally carry, pointing to a DCH DataConnection or an RHAI Connection. The `connection_ref` includes a type discriminator (`dch` or `rhai`) and the connection identifier (UUID for DCH, secret name for RHAI). In both cases, `connection_ref` references a *connection object* — not raw credentials. The actual secret is resolved through the respective connection API. This ADR defines the connection reference contract and three consumption scenarios for the 3.6 release.
```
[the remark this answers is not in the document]
abiazett: Not applicable — `3.6` is the correct RHOAI release version (the product uses a `major.minor` scheme starting from 1.x). The ADR text is accurate as written.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,310 @@
...
+
+DCH auto-migrates existing RHAI Connections by watching labeled Kubernetes secrets. When DCH and the Data Registry are both enabled, existing RHAI Connections are converted into DCH DataConnections. Data assets with a `connection_ref` pointing to an RHAI Connection continue to work — the underlying secret remains the same. New data assets created after migration will reference the DCH DataConnections directly.
+
+**Retrieving credentials from a DCH DataConnection:**
+
+For Scenario 1, users need direct access to connection credentials (e.g., S3 keys) without relying on DCH ingestion APIs. Rather than requiring users to mount the connection secret onto their pod — which can cause conflicts when multiple connections use the same environment variable names (e.g., `AWS_ACCESS_KEY_ID`) — DCH supports retrieving credentials programmatically:
+
+```
+GET /api/v1/data/connections/{id}?with_secret_info=true
+```
```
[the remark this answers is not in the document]
abiazett: Fixed — added `http` language identifier to the fenced code block.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,318 @@
...
+| Asset registered without `connection_ref` | Valid for local PVC volumes and assets where the user manages access independently. |
+| RHAI Connection after DCH auto-migration | Both `type: rhai` and `type: dch` references remain valid. The underlying Kubernetes secret is preserved by DCH during migration. |
+
+**Connection status visibility:** The Data Registry stores `connection_ref` as metadata — it does not track or validate the connection's runtime status. When an asset has a `connection_ref`, the consuming client (notebook, pipeline, agent, Data Hub UI) must call the DCH API (`GET /api/v1/data/connections/{id}`) to determine whether the connection exists, is ready, or is accessible to the current user. The Data Registry returns the asset regardless of connection state; it is the responsibility of the consuming client to check connection status and surface appropriate guidance — such as "connection not found", "connection not ready", or "insufficient permissions" — before the user attempts to access the data.
```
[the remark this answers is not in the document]
abiazett: Good catch — updated the "Connection status visibility" paragraph to route by connection type: `type: dch` checks via the DCH API, `type: rhai` checks the referenced Kubernetes secret directly in the asset's namespace.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,260 @@
...
+
+The Data Registry `connection_ref` field on tables and volumes stores a reference to a connection object — either a DCH DataConnection (identified by UUID) or an RHAI Connection (identified by secret name). The field is optional — assets with locally mounted PVC volumes do not require a connection reference.
+
+`connection_ref` always points to a connection *object*, never to raw credentials. It includes two subfields:
+
+| Subfield | Description | Example |
+|---|---|---|
+| `type` | Discriminator: `dch` or `rhai` | `dch` |
+| `id` | DCH DataConnection UUID or RHAI Connection secret name | `550e8400-e29b-41d4-a716-446655440000` |
+
```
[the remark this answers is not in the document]
abiazett: Out of scope for this ADR — `connection_ref` wire schema details (JSON Schema, validation rules, null vs omitted semantics, unknown type rejection) belong in the API specification, not the architecture decision record. The ADR defines the integration contract at the architectural level; the normative payload shape will be specified in the Data Registry API spec.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,260 @@
...
+- **`rhai`:** The `id` is the name of a labeled Kubernetes secret (`opendatahub.io/dashboard=true`, `opendatahub.io/managed=true`). The secret is resolved directly via the Kubernetes API.
+
+| Asset Type | Format | Connection Pattern | connection_ref |
+|---|---|---|---|
+| Table | iceberg | S3 credentials mounted on pod | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` |
+| Table | postgresql | DCH DataConnection (ingestion via DCH SDK) | `{type: dch, id: uuid}` |
+| Volume | s3 | S3 credentials mounted on pod | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` |
+| Volume | local (PVC) | OpenShift-managed, credentials on pod | None |
+
+DCH auto-migrates existing RHAI Connections by watching labeled Kubernetes secrets. When DCH creates a DCH DataConnection resource from an RHAI Connection, the asset's `connection_ref` can be updated from `{type: rhai, ...}` to `{type: dch, ...}` — but existing RHAI Connection references remain valid.
```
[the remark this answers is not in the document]
abiazett: Two sub-issues here:

1. **Label-based RHAI migration security** (watch permissions, label ownership, provenance checks) — @mariusdanciu can you check whether this is addressed in the DCH ADR ([PR #149](https://github.com/opendatahub-io/architecture-decision-records/pull/149))? This is DCH-internal behavior, not part of the integration contract defined here.

2. **connection_ref / asset format mismatch validation** — valid point. Adding an edge case to the 3.6 Edge Cases table specifying that mismatched connection types should be rejected.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,260 @@
...
+
+    User->>DR: GET /catalog/projects/{p}/collections/{c}/tables/claims-db
+    DR-->>User: table metadata<br/>format: postgresql<br/>connection_ref: {type: dch,<br/>id: 550e8400-...}<br/>schema: [claim_id, status, amount]
+
+    User->>DCH: dch.ingest(connection_id="550e8400-...",<br/>query="SELECT * FROM claims")
+    Note over User,DCH: Headers: Authorization, x-tenant-id,<br/>x-dch-connection-id
+
+    DCH->>PG: Connect using stored credentials<br/>Execute query
+    PG-->>DCH: Result rows
+    DCH-->>User: Arrow RecordBatch
```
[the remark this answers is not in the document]
abiazett: DCH-internal authorization design — not part of the integration contract defined in this ADR. @mariusdanciu can you check whether the DCH ADR ([PR #149](https://github.com/opendatahub-io/architecture-decision-records/pull/149)) addresses bearer token propagation, `x-tenant-id` derivation from authenticated identity, and fail-closed SAR before secret reads?

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+The user never sees database credentials. DCH handles driver dependencies, connection pooling, and credential management. The Data Registry provides the `connection_ref` that links the table asset to the right DCH DataConnection.
```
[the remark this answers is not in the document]
abiazett: DCH-internal implementation concern — SQL query scoping, read-only database roles, and tenant/PII authorization on `dch.ingest` are DCH responsibilities, not part of this integration contract. @mariusdanciu can you check whether the DCH ADR ([PR #149](https://github.com/opendatahub-io/architecture-decision-records/pull/149)) covers query scope enforcement and read-only constraints?

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,301 @@
...
+
+| Scenario | Data Registry Permission | DCH Permission |
+|---|---|---|
+| Scenario 0: Discover assets | `dataregistry-view` in namespace | -- |
+| Scenario 1: Iceberg table (mounted secret) | `dataregistry-view` in namespace | -- |
+| Scenario 2a: DCH streaming (Arrow Flight) | `dataregistry-view` in namespace | `data-connection: read` + `data-store: get` |
+| Scenario 2b: DCH ExportedSecret (direct access) | `dataregistry-view` in namespace | `data-connection: read` + K8s Secret RBAC for ExportedSecret |
+| Scenario 3: PVC volume (no connection) | `dataregistry-view` in namespace | -- |
+
+Both components use ClusterRole aggregation labels to auto-inject permissions into standard OpenShift roles (`view`, `edit`, `admin`). Users with existing namespace roles automatically receive the corresponding Data Registry and DCH permissions without additional RBAC configuration.
```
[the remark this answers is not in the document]
abiazett: Not applicable — namespace `view` must retain access to DCH connections because RHAI Connections work this way today: if you have `view` access on a project/namespace, you can view the RHAI Connections in that namespace. Restricting DCH connection visibility to a separate consumer role would break existing clients. @mariusdanciu FYI.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,310 @@
...
+
+**Retrieving credentials from a DCH DataConnection:**
+
+For Scenario 1, users need direct access to connection credentials (e.g., S3 keys) without relying on DCH ingestion APIs. Rather than requiring users to mount the connection secret onto their pod — which can cause conflicts when multiple connections use the same environment variable names (e.g., `AWS_ACCESS_KEY_ID`) — DCH supports retrieving credentials programmatically:
+
+```
+GET /api/v1/data/connections/{id}?with_secret_info=true
+```
+
+By default, the DCH DataConnection response does not include credential values — the admin/secret section is omitted for regular users. If the caller has RBAC permissions to read secrets (`secrets: get` in the namespace), the `?with_secret_info=true` query parameter instructs DCH to unpack the secret keys and include them in the response. This allows users to retrieve credentials on demand without pod-level secret mounting, avoiding environment variable collisions across multiple connections.
```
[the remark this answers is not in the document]
abiazett: Not applicable — namespace-wide `secrets:get` is how RHAI Connections work today: if you have `view` access on a project/namespace, you can view the RHAI Connections in that namespace. Restricting this to a DCH-scoped credential permission would break existing clients. Note: this is not credential vending — DCH returns credentials from Kubernetes secrets given proper RBAC permissions, it does not integrate with cloud-native credential resolution services. @mariusdanciu FYI.

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,318 @@
...
+- **`rhai`:** The `id` is the name of a labeled Kubernetes secret (`opendatahub.io/dashboard=true`, `opendatahub.io/managed=true`). The secret is resolved directly via the Kubernetes API.
+
+In the table below we list the different Data Registry Asset Types and formats, and which connection pattern they support as well as whether DCH will support their direct Ingestion (in 3.6).
+
+| Asset Type | Format | Connection Pattern | connection_ref | Ingestion |
+|---|---|---|---|---|
+| Table | iceberg | S3 credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (PyIceberg, Spark, Trino) |
+| Table | postgresql | DCH DataConnection | `{type: dch, id: uuid}` | DCH SDK |
+| Table | oci (OCI Artifact registry) | Credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | Direct (user libraries) |
+| Volume | s3 | S3 credentials via DCH API or RHAI secret | `{type: dch, id: uuid}` or `{type: rhai, id: secret-name}` | DCH SDK (jsonl, binary via Flight) or Direct (user libraries) |
```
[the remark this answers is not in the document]
abiazett: DCH API specification detail — the S3 Flight operation contract (object binding, format negotiation, streaming, size limits, error handling) belongs in the DCH API spec, not this integration ADR. @mariusdanciu can you check whether the DCH ADR ([PR #149](https://github.com/opendatahub-io/architecture-decision-records/pull/149)) covers this?

architecture-decision-records/data-registry/ODH-ADR-DR-0002-data-registry-dch-integration.md
```diff
@@ -0,0 +1,318 @@
...
+    participant User as Data Scientist<br/>(Workbench Pod)
+    participant DR as Data Registry API
+    participant DCH as DCH Service<br/>(Arrow Flight)
+    participant PG as PostgreSQL
+
+    User->>DR: GET /v1/{p}/namespaces/{c}/generic-tables/auto-claims-db
+    DR-->>User: table metadata<br/>format: postgresql<br/>connection_ref: {type: dch,<br/>id: 550e8400-...}<br/>schema: [claim_id, status, amount]
+
+    User->>DCH: dch.ingest(connection_id="550e8400-...",<br/>query="SELECT * FROM claims")
+    Note over User,DCH: Headers: Authorization, x-tenant-id,<br/>x-data-connection-id
```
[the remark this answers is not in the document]
abiazett: DCH-internal request validation — the relationship between `connection_id` in the SDK call and `x-data-connection-id` in the headers is a DCH API design concern, not part of this integration contract. @mariusdanciu can you check whether the DCH ADR ([PR #149](https://github.com/opendatahub-io/architecture-decision-records/pull/149)) defines the canonical identifier and mismatch rejection behavior?

PR #155, opened 2026-08-17
ADR-MS-0005: retain deprecated kserve.modelsAsService field

lburgazzoli:
Documents the decision to keep spec.components.kserve.modelsAsService
in the DSC v2 API (defaulted to Removed, one-directional CEL) instead
of removing/relocating it or adding a new enum value, following the
MaaS move to the AI Gateway module.

## Merge criteria:

## Summary by CodeRabbit

* **Documentation**
  * Added an architectural decision record documenting continued schema compatibility for the deprecated model-serving configuration field.
  * Clarified compatibility behavior across API versions, including conversion of legacy values, stored-value retention, defaulting, fallback behavior, and precedence rules.
  * Documented that the deprecated field permits only the transition from Managed to Removed and produces a warning when set to Managed.

### DIFF
--- a/architecture-decision-records/model-serving/ODH-ADR-MS-0005-retain-deprecated-modelsasservice-field.md
+++ b/architecture-decision-records/model-serving/ODH-ADR-MS-0005-retain-deprecated-modelsasservice-field.md
@@ -1,0 +1,100 @@
+# ODH-ADR-MS-0005: Retain deprecated `kserve.modelsAsService` field
+
+|                |            |
+| -------------- | ---------- |
+| Date           | 2026-08-17 |
+| Scope          | Model Serving, Models-as-a-Service (MaaS), AI Gateway |
+| Status         | Draft |
+| Authors        | Luca Burgazzoli |
+| Supersedes     | N/A |
+| Superseded by: | N/A |
+| Tickets        | [RHOAIENG-60628](https://redhat.atlassian.net/browse/RHOAIENG-60628), [RHOAIENG-76610](https://redhat.atlassian.net/browse/RHOAIENG-76610) |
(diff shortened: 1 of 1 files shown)


somya-bhatnagar:
Thanks for documenting this decision. One clarification would help future readers understand how the fallback actually works in the operator:

**In the How section**, the line "aigateway.modelsAsAService.managementState is the canonical MaaS toggle and takes precedence whenever it's set" conflates two independent fallback paths that operate at different layers:

1. **Module on/off** (`PopulatePlatformModule`): Falls back on **empty** `aigateway.managementState` (not the nested field). If `aigateway.managementState` is explicitly `Removed`, the deprecated field is ignored even if still `Managed`. Fallback still requires `kserve.managementState == Managed`.

2. **Nested MaaS config** (`BuildModuleCR`): Falls back on empty `aigateway.modelsAsAService.managementState`, and still requires `kserve.managementState == Managed`.

A truth table would help:

| Fresh 3.5 | Stored v2, old `Managed` | v1 write | Both set | Parent `Removed` + old `Managed` |
|-----------|-------------------------|---------|---------|-------------------------------------|
| `aigateway.managementState: Managed` | Fallback → `Managed` | ConvertTo → v2, Warning | Explicit wins | Explicit `Removed` wins |

This matters because the migration guide already notes ("Use Removed, not omit") that empty `aigateway.managementState` re-triggers the legacy path — so "whenever it's set" should clarify that "set" means non-empty, not "explicitly written by the user."

Minor: "now defaults to Removed instead of Managed" — the field already defaulted to `Removed` in 3.4 (no change there).

lphiri (dismissed):
LGTM

lphiri (approved):

PR #156, opened 2026-08-19
ODH-ADR-Operator-0015: RHOAI multi-tenancy framework (draft)

lphiri:
## Summary

Draft ADR proposing a top-level, platform-wide multi-tenancy framework for RHOAI. Opening as a draft to get **early feedback** on direction before it is finalized.

The framework introduces a first-class tenancy control plane under a new API group `tenancy.opendatahub.io/v1alpha1`, delivered as a controller plus a fail-closed validating webhook, built around three cluster-scoped CRDs:

- **`PlatformTenant`** — allocation (capacity, quota, hardware, hierarchy); parent-controlled.
- **`TenantProfile`** — configuration (admins, isolation, defaults, observability); self-managed.
- **`TenantProject`** — namespace provisioning (Namespace/Project + LocalQueue, RBAC, NetworkPolicies).

### Key decisions

- **Kueue-native quota**: parent -> Cohort, leaf -> ClusterQueue, project -> LocalQueue; capacity is a control-plane budget enforced by the webhook. RHBoK is a required dependency, provisioned out of band.
- **Delegated MaaS provisioning**: the framework creates and owns an `AITenant`; the existing maas-controller reconciles it unchanged as the delegated provisioner (realizes the ODH-ADR-MS-0003 extension point). No MaaS schema or reconciliation changes.
- **Tiered observability** (labeling / Perses dashboards / dedicated MonitoringStack + OTel filtering relay) as the showback/chargeback substrate.
- **Opt-in and backward compatible**: `Tenancy` component off by default; existing namespaces, ClusterQueues, and hand-created AITenants keep working; adoption is admin-initiated.

### Suggested execution plan

Sequenced by customer value: Phase 1 tenancy core for GPUaaS -> **Phase 2 observability for GPUaaS (top priority)** -> Phase 3 MaaS integration -> Phase 4 advanced isolation and extensions.

### Files

- `architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md`
- `architecture-decision-records/operator/multi-tenancy-strategy.md` (supporting strategy)

### Open questions for reviewers

- Packaging: embedded platform service vs standalone operator.
- Hierarchy depth safety guard and sensible default.
- MaaS intent placement/scope (on `TenantProfile.spec.services.maas` vs a separate CR; leaf/standalone-only vs parent-level shared gateway).
- Scalability targets beyond the 50-tenant / 200-project design target.

Tickets: RHAIRFE-2922, RHAISTRAT-2554 (source RFE RHAIRFE-3115)

### DIFF
--- a/architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
+++ b/architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
@@ -2,115 +2,115 @@
 
 |                |            |
 | -------------- | ---------- |
-| Date           | 2026-08-18 |
+| Date           | 2026-08-19 |
 | Scope          | Operator (RHOAI platform-wide multi-tenancy) |
 | Status         | Draft |
-| Authors        | [Lindani Phiri](@lphiri) |
+| Authors        | [Lindani Phiri](@lphiri) |  [Chris Sams](@csams)
 | Supersedes     | N/A |
 | Superseded by: | N/A |
 | Tickets        | [RHAIRFE-2922](https://redhat.atlassian.net/browse/RHAIRFE-2922), [RHAISTRAT-2554](https://redhat.atlassian.net/browse/RHAISTRAT-2554) |
-| Other docs:    | operator/multi-tenancy-strategy.md; Tenancy Control Plane design (multitenant-gpuaas: PLAN.md, API-REFERENCE.md, KUEUE-API-REFERENCE.md, ADRs 0001-0033); [RHAIRFE-3115](https://redhat.atlassian.net/browse/RHAIRFE-3115) |
+| Other docs:    | operator/multi-tenancy-strategy.md; Tenancy Control Plane design (multitenant-gpuaas: PLAN.md, API-REFERENCE.md); [RHAIRFE-3115](https://redhat.atlassian.net/browse/RHAIRFE-3115) |
 
 ## What
 
 RHOAI will introduce a top-level multi-tenancy framework: a first-class tenancy
-control plane that represents, manages, and enforces isolation boundaries for
-shared GPU/AI clusters. The framework is delivered as a controller plus
+control plane that represents an organizational tenant hierarchy and provides
+governed, self-service namespace provisioning with delegated administration,
+network isolation, per-tenant observability, and delegated Models-as-a-Service
+(MaaS) provisioning. The framework is delivered as a controller plus a
 validating webhook under a new API group `tenancy.opendatahub.io/v1alpha1`, and
 is built around three cluster-scoped custom resources:
 
-* **`PlatformTenant`** — the allocation resource. Parent-controlled. Defines
-  resource budget, hardware scope, and position in the tenant hierarchy.
+* **`PlatformTenant`** — the hierarchy resource. Parent-controlled. Defines a
+  tenant's existence and its position in the organizational tree.
 * **`TenantProfile`** — the configuration resource. Self-managed by tenant
-  admins. Holds admins, isolation policy, sharing preferences, defaults, and
-  observability.
+  admins. Holds admins, defaults, network and metadata policy, observability,
+  and optional service bindings (such as MaaS).
 * **`TenantProject`** — the namespace provisioning resource. Maps 1:1 to a
-  Namespace (or OpenShift Project) plus a Kueue LocalQueue, with RBAC and
-  NetworkPolicies.
+  Namespace (or OpenShift Project) with RBAC and NetworkPolicies.
 
-Rather than inventing a bespoke quota engine, the framework maps the tenant
-hierarchy onto Kueue: parent tenants become Kueue Cohorts, leaf tenants become
-ClusterQueues, and projects become LocalQueues. Quota, borrowing, and
-preemption are enforced by Kueue at runtime; the control plane enforces the
-administrative allocation budget (capacity) through the webhook. Red Hat Build of
-Kueue (RHBoK) is a required dependency, provisioned out of band (not through OLM
-dependency resolution) and assumed present before the framework reconciles. The
-remaining capabilities activate by CRD detection (DRA, OpenShift, COO, OTel,
-Perses), so the control plane degrades gracefully for those and can run on
-vanilla Kubernetes as well as OpenShift.
+The framework models the tenant hierarchy as a plain tree: a `PlatformTenant`
+may have child `PlatformTenant`s, `TenantProject`s, or both. It provides the
+tenant identity, delegation, and isolation substrate. Resource quota and
+fairness (for example a Kueue-based GPU allocation model) are deliberately out of
+scope here and deferred to a future ADR that can build on this tenant identity.
+Capabilities activate by CRD detection (OpenShift, COO, OTel, Perses), so the
+control plane degrades gracefully and runs on both OpenShift and vanilla
+Kubernetes.
 
 ## Why
 
 Today, onboarding a team onto a shared RHOAI cluster means manually stitching
-together namespace creation, RBAC role bindings, resource quotas, Kueue
-ClusterQueues, network policies, and storage/ingress configuration. This is
-slow, error-prone, and inconsistent across tenants. The operational burden
-grows linearly with each new team, and the risk of security misconfiguration
-grows with it. Prior analysis of the GPUaaS backlog found six different working
-definitions of "multi-tenancy" and four competing primitives each claiming to
-be "the tenant." Without a single, first-class tenant abstraction, every large
-deployment reinvents tenancy from raw primitives.
+together namespace creation, RBAC role bindings, network policies, and
+storage/ingress configuration. This is slow, error-prone, and inconsistent
+across tenants. The operational burden grows linearly with each new team, and
+the risk of security misconfiguration grows with it. Prior analysis of the
+backlog found six different working definitions of "multi-tenancy" and four
+competing primitives each claiming to be "the tenant." Without a single,
+first-class tenant abstraction, every large deployment reinvents tenancy from raw
+primitives.
 
 Enterprise and sovereign-AI customers deploying RHOAI for multiple teams are
-already raising escalations about RBAC drift, inconsistent quota enforcement,
-and the difficulty of proving consistent isolation during compliance audits.
+already raising escalations about RBAC drift, inconsistent isolation, and the
+difficulty of proving consistent isolation during compliance audits.
 
 The driving forces are best framed by the multi-tenancy strategy
-(`operator/multi-tenancy-strategy.md`), whose statement is to *build a multi-tenancy framework
-that makes shared Kubernetes simple to consume, safe to operate, and
-economically transparent*. That strategy treats multi-tenancy as a customer
-operating model, not merely a security architecture. Its pillars map directly to
+(`operator/multi-tenancy-strategy.md`), whose statement is to *build a
+multi-tenancy framework that makes shared Kubernetes simple to consume, safe to
+operate, and economically transparent*. That strategy treats multi-tenancy as a
+customer operating model, not merely a security architecture. Its pillars map to
 what this framework delivers:
 
-* **Tenant lifecycle:** declarative create/update/decommission through three CRs
+* **Tenant lifecycle:** declarative create/update/decommission through the CRs
   instead of tickets and runbooks.
-* **Isolation and governance:** the secure path is the default path;
-  auto-created `TenantProfile` starts maximally restrictive (hard isolation,
-  strict network, no projects) until an admin opens it up.
-* **Self-service management:** delegated RBAC lets a tenant admin subdivide
-  their own allocation and provision projects without platform-admin
-  involvement.
-* **Resource fairness:** Kueue Cohorts/ClusterQueues express the shared-resource
-  contract; isolation presets (hard/guaranteed/balanced/burst) tune
-  borrowing, lending, and preemption.
+* **Isolation and governance:** the secure path is the default path; an
+  auto-created `TenantProfile` starts maximally restrictive (strict network, no
+  projects) until an admin opens it up.
+* **Self-service management:** delegated RBAC lets a tenant admin subdivide their
+  own subtree and provision projects without platform-admin involvement.
 * **Cost attribution:** always-on metric labeling plus optional per-tenant
   dashboards and dedicated monitoring provide the consumption substrate for
   showback and chargeback.
 
-Without a structured tenant model, RHOAI cannot serve as a shared platform at
-enterprise scale and loses ground to AI platforms that ship multi-tenant
-governance out of the box.
+The strategy's **resource fairness** pillar (quota, limits, priority) is
+acknowledged but deferred to a follow-up ADR so this foundational tenant model
+can land first. Without a structured tenant model, RHOAI cannot serve as a
+shared platform at enterprise scale and loses ground to AI platforms that ship
+multi-tenant governance out of the box.
 
 ## Goals
 
-* Provide a single, first-class tenant abstraction over Kubernetes namespaces
-  and Kueue quota, replacing ad-hoc assembly from raw primitives.
-* Split allocation (parent-controlled) from configuration (self-managed) into
-  `PlatformTenant` and `TenantProfile` so RBAC can be scoped per resource and a
-  tenant admin cannot escalate their own quota.
-* Map the hierarchy onto Kueue Cohort trees, delegating runtime quota,
-  borrowing, and preemption to Kueue rather than a bespoke engine.
-* Enforce the delegation invariant: the sum of a parent's children's
-  allocations can never exceed the parent's capacity, per resource type.
+* Provide a single, first-class tenant abstraction over Kubernetes namespaces,
+  replacing ad-hoc assembly from raw primitives.
+* Model an organizational tenant hierarchy (a tree of `PlatformTenant`s) that
+  supports delegated administration down the tree.
+* Split hierarchy/existence (parent-controlled) from configuration
+  (self-managed) into `PlatformTenant` and `TenantProfile` so RBAC can be scoped
+  per resource and a tenant admin cannot change their own placement in the tree.
 * Enable delegated self-service: a tenant admin creates sub-tenants and projects
-  within their assigned budget without platform-admin intervention.
-* Offer named isolation presets for quota sharing (hard/guaranteed/balanced/
-  burst) and network isolation (none/tenant/strict), with a raw override escape
-  hatch.
+  within their subtree without platform-admin intervention.
+* Provide network isolation presets (none/tenant/strict) with cross-tenant
+  grants.
 * Provide tiered observability: always-on metric labeling, opt-in Perses
   dashboards, and opt-in per-tenant dedicated monitoring for showback/chargeback.
+* Provide delegated MaaS tenant provisioning built on the tenant identity.
 * Activate capabilities by CRD detection so the control plane degrades
   gracefully and runs on both OpenShift and vanilla Kubernetes.
 * Preserve full backward compatibility: unmanaged namespaces operate unchanged;
-  adoption of existing namespaces and Kueue resources is opt-in.
+  adoption of existing namespaces is opt-in.
 
 ## Non-Goals
 
+* Resource quota and fairness. GPU/compute quota, borrowing/preemption, and
+  Kubernetes ResourceQuota are out of scope for this ADR and deferred to a
+  follow-up (for example a Kueue-based allocation model layered on this tenant
+  identity). This framework provides the tenant hierarchy those systems attach
+  to; it does not allocate or enforce resources.
 * Multi-cluster tenancy federation. Single-cluster scope; cross-cluster is
-  deferred to a future MultiKueue-based design.
+  deferred to a future design.
 * End-user self-service provisioning of arbitrary root tenants. Root
   `PlatformTenant` creation remains a cluster-admin function; self-service is
-  limited to delegated subdivision within an assigned budget (this reconciles
+  limited to delegated subdivision within an assigned subtree (this reconciles
   the two source tickets; see Alternatives).
 * Billing and chargeback systems and FinOps reporting UIs. This framework
   delivers the consumption/metrics substrate; chargeback tooling is downstream.
@@ -120,15 +120,11 @@ governance out of the box.
   maas-controller reconciles unchanged (see How: Delegated MaaS tenant
   provisioning); it does not alter MaaS internals.
 * Per-tenant filtering of shared cluster-scoped catalogs (HardwareProfile,
-  ClusterServingRuntime, WorkspaceKind, etc.) beyond DRA DeviceClass gating.
-  A candidate future extension.
-* StorageClass access gating (deferred past v1).
-* Automated hierarchy re-parenting or role transitions. A tenant's role
-  (parent/leaf/standalone) is immutable; reorganizing is done by creating new
-  nodes and migrating.
-* Shipping Kueue itself. RHBoK is a required dependency the framework consumes,
-  provisioned out of band; this ADR does not change how Kueue is packaged,
-  shipped, or installed.
+  ClusterServingRuntime, WorkspaceKind, etc.). A candidate future extension.
+* Hardware/accelerator access gating (DeviceClass, StorageClass). Deferred, and
+  naturally paired with the future quota ADR.
+* Automated hierarchy re-parenting. A tenant's parent is immutable; reorganizing
+  is done by creating new nodes and migrating.
 
 ## How
 
@@ -141,67 +137,75 @@ namespace inside a tenant.
 `PlatformTenant` is named `PlatformTenant` (not `Tenant`) deliberately, to avoid
 collision with the MaaS `maas.opendatahub.io Tenant`/`AITenant` CRDs.
 
-A `PlatformTenant` is one of three roles, fixed at creation and immutable:
+The split exists for per-resource RBAC. `PlatformTenant` records a tenant's
+existence and its position in the tree (`spec.parent`); only parent admins (or
+cluster-admin for roots) may create or delete it, so a tenant admin cannot
+re-parent or fabricate tenants. `TenantProfile` is auto-created 1:1 with
+maximally restrictive defaults (empty admins, `strict` network, `maxProjects: 0`)
+and is then self-managed by the tenant admins the parent bootstraps.
+`TenantProject` maps 1:1 to a namespace and generates RBAC and NetworkPolicies.
 
-| Role | Parent? | Children? | Projects? | Generates | Purpose |
-|---|---|---|---|---|---|
-| Parent (branch) | optional | yes | no | Cohort, TenantProfile | Sharing boundary, capacity budget, hardware scope |
-| Leaf | yes | no | yes | ClusterQueue (in parent's Cohort), TenantProfile | Quota holder, workload admission |
-| Standalone | no | no | yes | ClusterQueue (no Cohort), TenantProfile | Simple single-tenant, no hierarchy |
+### Tenant hierarchy
 
-`TenantProfile` is auto-created by the controller with maximally restrictive
-defaults (empty admins, `hard` isolation, no borrow/lend, `strict` network,
-`maxProjects: 0`). The parent admin bootstraps the initial `admins`; from then
-on those admins self-manage the profile. `TenantProject` maps 1:1 to a namespace
-plus a LocalQueue and generates RBAC, NetworkPolicies, and (optionally) a
-ResourceQuota.
+The hierarchy is a plain organizational tree. A `PlatformTenant` may have child
+`PlatformTenant`s, `TenantProject`s, or both. There is no fixed depth: the common
+case is a single tenant (one level) or a division with team sub-tenants (two
+levels), and deeper nesting is available but never required. The tree exists to
+support delegated administration, subtree-scoped observability, and consistent
+tenant/cost-attribution labeling, not resource allocation.
 
-### Mapping to Kueue and the capacity model
+`spec.parent` is immutable. An operator-level setting may impose a conservative
+maximum depth as a safety guard, but the framework does not promote deep trees.
 
-The hierarchy maps onto Kueue primitives:
+### Delegated administration and authorization
 
-* Parent `PlatformTenant` -> Cohort (nested under the parent's Cohort).
-* Leaf `PlatformTenant` -> ClusterQueue in the parent's Cohort, with
-  `nominalQuota = guaranteed`.
-* `TenantProject` -> Namespace + LocalQueue pointing at the tenant's CQ.
+Administration is delegated down the tree. A parent admin creates a child
+`PlatformTenant` and bootstraps its initial `TenantProfile.spec.admins`; from
+then on those tenant admins self-manage the profile and create projects and
+sub-tenants within their subtree, up to `TenantProfile.spec.defaults.maxProjects`.
 
-**Capacity** is a control-plane concept that does not exist in Kueue: the
-maximum total allocation a parent allows its children to divide. It is enforced
-at admission by the webhook (`sum(children allocations) <= parent capacity` per
-resource type), not by Kueue quota. All generated Cohorts use `nominalQuota = 0`;
-sharing happens through CQ-level lending of unused guaranteed quota. This keeps
-capacity a clean per-resource-type API concept and avoids Kueue's per-(flavor,
-resource) distribution problem.
+Admin authority is scoped per layer rather than inherited wholesale. Making a
+parent admin an automatic admin of everything below them would collapse the
+`PlatformTenant`/`TenantProfile` split and maximize blast radius, so inheritance
+applies only where subtree ownership genuinely requires it:
 
-Delegation is protected by four layers: (1) RBAC grants tenant admins CRUD on
-`TenantProfile`/`TenantProject` and create-only on child `PlatformTenant`, but
-no access to Kueue CRs; (2) the validating webhook enforces the capacity ceiling
-and hardware-subset rules on every create/update; (3) managed Kueue resources
-are protected from direct edits; (4) Kueue enforces runtime usage against the
-quotas the controller derived from the validated spec.
+* **Existence, placement, and admin assignment - inherited transitively.** Any
+  *ancestor* admin (not only the immediate parent) may create or delete
+  descendant `PlatformTenant`s and may write a descendant's
+  `TenantProfile.spec.admins`. This is what owning a subtree means, and it is the
+  orphan-recovery path: if a tenant's admins all leave, an ancestor (ultimately
+  cluster-admin) reassigns them. Because this authority is computed from the tree,
+  not from being listed in the child's `spec.admins`, a tenant admin cannot lock
+  an ancestor out.
+* **Configuration self-management - not inherited.** Only a tenant's own
+  `spec.admins` may edit the rest of its `TenantProfile` (network defaults,
+  observability, `services`) and its `TenantProject`s. An ancestor can still take
+  over, but only by first rewriting `spec.admins`, which is an explicit, audited
+  act rather than silent ambient access.
+* **Namespace and data access - never inherited.** Access to a project's
+  namespace comes solely from `TenantProject.spec.users`; ancestor admins get no
+  automatic `kubectl` or data access to descendant workloads.
 
-### Hierarchy depth
+Delegation is protected by three layers:
 
-Depth is not a fixed 5-level generic tree. It emerges naturally from Kueue
-Cohort nesting, and the parent/leaf/standalone role model keeps it manageable:
-the common case is a **standalone** tenant (one level) or a **single parent with
-leaf children** (two levels). Deeper nesting is available for organizations that
-need it, but is never required and is not the default mental model. An
-operator-level setting may impose a conservative maximum depth as a safety
-guard, but the framework does not promote deep trees. This deliberately avoids
-the complexity of the previously proposed "configurable, capped at 5 levels"
-model.
+1. **RBAC** grants tenant admins CRUD on `TenantProfile`/`TenantProject` and
+   create-only on child `PlatformTenant`.
+2. **The validating webhook** is the real authorization gate. It is fail-closed
+   (`failurePolicy: Fail`) and checks, on every mutating operation, that the
+   requesting user is authorized for that layer: an ancestor admin for
+   `PlatformTenant` create/delete and for edits to `TenantProfile.spec.admins`;
+   the tenant's own admins for the rest of `TenantProfile` and for
+   `TenantProject`. "Is caller an ancestor admin?" is a bounded walk up
+   `spec.parent` (capped by the depth guard) checking membership in each
+   ancestor's `spec.admins`. Child creation is additionally verified with a
+   SubjectAccessReview against the parent.
+3. **Managed-resource protection** prevents direct edits to controller-owned
+   resources (namespaces, RBAC, NetworkPolicies) outside the tenancy API.
 
-### Isolation presets
-
-Named presets configure the generated ClusterQueue (leaf) or Cohort (parent)
-borrowing/lending limits and preemption policy, with a raw Kueue override
-available on `TenantProfile.spec.overrides`:
-
-* `hard` — fully isolated; no borrowing or lending.
-* `guaranteed` — donates unused quota to siblings but never borrows.
-* `balanced` — full sharing within the Cohort (Kueue default).
-* `burst` — hoards its own quota but can borrow up to a cap for peak demand.
+Kubernetes RBAC cannot express "only the tenants you own," so the
+`tenancy:tenant-admin` ClusterRole is intentionally broad and the webhook does
+the scoping. `PlatformTenant` placement is parent-owned, so a tenant admin cannot
+escalate by editing their own hierarchy node.
 
 ### Network isolation
 
@@ -209,13 +213,13 @@ Per-`TenantProject` `networkIsolation` generates baseline NetworkPolicies:
 
 * `none` — no policies created.
 * `tenant` — blocks cross-tenant in-cluster lateral movement; allows external
-  egress (S3, registries, cloud APIs) and same-leaf-tenant traffic.
+  egress (S3, registries, cloud APIs) and same-tenant traffic.
 * `strict` — full lockdown; only system namespaces and DNS allowed.
 
-System namespaces are auto-discovered from DSCI, Kueue, Gateway/Istio, and
-Kuadrant and labeled for selector-based policies. Component-created
-NetworkPolicies layer additively; the controller never deletes policies it does
-not own. Cross-namespace `networkGrants` create matching rules on both sides.
+System namespaces are auto-discovered from DSCI, Gateway/Istio, and Kuadrant and
+labeled for selector-based policies. Component-created NetworkPolicies layer
+additively; the controller never deletes policies it does not own. Cross-tenant
+`networkGrants` create matching rules on both sides.
 
 ### Observability and cost attribution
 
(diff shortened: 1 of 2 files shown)


architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
...
+| Predictable performance | Resource quotas, limit ranges, priority controls, capacity policies | Fewer noisy-neighbor incidents and clearer service expectations |
+| Operational simplicity | Standard observability, logs, dashboards, alerts, audit trails, platform support model | Less Kubernetes expertise required from application teams |
+| Cost accountability | Mandatory ownership metadata, usage metering, allocation rules, showback and chargeback reports | Teams understand what they consume and can manage their spend |
+| Business-level economics | Cost-to-serve by tenant, product, environment, or customer | Finance and product leaders can assess margin, pricing, and investment decisions |
+
+Grouping tenant objects, restricting tenant access, controlling resource consumption, isolating network traffic, and enforcing baseline deployment standards are core multi-tenancy capabilities. Cost allocation should then assign shared CPU, memory, storage, network, and other infrastructure costs proportionally to the tenants that consume them.
```
mariusdanciu, line 20: Cost attribution seem to be part of the ADR although the RHAISTRAT-2554 seems to treat cost attribution as out-of scope.

architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
...
+## Cost model
+
+Use a transparent layered model:
+
+1. **Direct costs:** Attribute compute, memory, GPU, storage, network, and managed-service consumption directly to the workload or tenant where possible.
+2. **Shared cluster costs:** Allocate worker-node, control-plane, observability, ingress, and shared storage costs using an agreed driver—such as actual consumption, requested capacity, or a blended approach.
+3. **Unallocated costs:** Report resources without valid ownership metadata separately and assign them temporarily to a platform or shared-services cost center.
+4. **Business unit economics:** Roll infrastructure costs into cost per tenant, product, transaction, API call, workload, or customer—depending on the business model.
+
+This model should make room for both **showback**—cost visibility without financial enforcement—and **chargeback**—allocating costs to accountable budgets. Over time, unit-cost metrics can connect technical consumption to customer economics, such as cost per transaction or cost per active customer.
```
mariusdanciu, line 84: Who can view what costs ? Is it correct to assume that a TU admin subject can view the costs only for its TenancyUNit and its child TenancyUnits, nut never parentRefs ?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,1080 @@
...
+collision with the MaaS `maas.opendatahub.io Tenant`/`AITenant` CRDs.
+
+A `PlatformTenant` is one of three roles, fixed at creation and immutable:
+
+| Role | Parent? | Children? | Projects? | Generates | Purpose |
+|---|---|---|---|---|---|
+| Parent (branch) | optional | yes | no | Cohort, TenantProfile | Sharing boundary, capacity budget, hardware scope |
+| Leaf | yes | no | yes | ClusterQueue (in parent's Cohort), TenantProfile | Quota holder, workload admission |
+| Standalone | no | no | yes | ClusterQueue (no Cohort), TenantProfile | Simple single-tenant, no hierarchy |
+
```
lburgazzoli: The `PlatformTenant` having different roles depending on the spec shape is a little bit confusing and add some cognitive load. It amy also cause quite some overhead in any webhook since the webhook should eventually figure out the actual CR role.

I'd suggest we have a dedicated CRD for the leaf and grouping one like:
- `Tenant`: leaf or standalone, maps to `ClusterQueue`
- `TenantGroup` : maps to `Cohort`

I'm aware that `Tenant` kind clashes with the on of the MaaS type (but it is still distinct as the types belong to different api groups), but I would eventually rename/remove the MaaS one in the long term.

architecture-decision-records/operator/multi-tenancy-strategy.md
csams: Yeah, the idea is a tenant should be able to see its subtree but not up the tree.

The other idea is a root tenant gets its own metrics store (or some kind of isolation guarantee) for this data.

architecture-decision-records/operator/multi-tenancy-strategy.md
csams: We should stay away from cost attribution.

architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
+A Kubernetes multi-tenancy framework should be positioned as a **customer operating model**, not merely a security architecture. Customers need to safely share infrastructure, but they also need a low-friction way to onboard teams, apply consistent policies, see consumption, and attribute spend to the right customer, product, team, or cost center. Kubernetes cost allocation requires connecting cloud billing, cluster usage, and workload metadata such as namespace and labels.
```
etirelli, line 1: This strategy defines a tenant as a set of labeled namespaces, and tenant identity is inherited from where a workload runs. This works for workload-bound resources (GPU, compute, pipeline runs, inference) as these resources run inside tenant namespaces. However, it does not cover RHOAI's shared platform services — MLflow, Feast, AI Hub, Dashboard, EvalHub — which run as cluster-wide outside any tenant namespace. For these the model provides neither attribution (consumption is labeled to the platform namespace, not a tenant) nor isolation (tenants share one instance and see each other's data unless the service enforces otherwise (e.g. mlflow supports soft tenancy isolation)). 

I suggest adding a section that maps each of the platform's major components with a classification (per-tenant instance, shared tenant-aware instance and shared singleton (global) cross-tenant instance). Then for each group we need a plan for how tenancy is either enforced or deferred.

architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
...
+- Environment, such as production, staging, and development
+- Cluster, region, and infrastructure type
+- Workload and namespace
+- Shared platform services, allocated through a published methodology
+
+Labels for team, environment, and application, combined with namespaces that reflect business or environmental boundaries, are common building blocks for actionable Kubernetes cost allocation.
+
+## Cost model
+
+Use a transparent layered model:
```
etirelli, line 77: I suggest explicitly stating how shared services fit into the cost model. Some might be able to attribute direct costs (mlflow experiments/scorers/judges), vs some might have to assign unnalocated costs (dashboard?)

architecture-decision-records/operator/multi-tenancy-strategy.md
lphiri: Thanks for the feedback  @etirelli  will refine this to clarify.

architecture-decision-records/operator/multi-tenancy-strategy.md
lphiri: Agreed,  the ADR does call this as out of scope, but will clarify here as well.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,1080 @@
...
+|                |            |
+| -------------- | ---------- |
+| Date           | 2026-08-18 |
+| Scope          | Operator (RHOAI platform-wide multi-tenancy) |
+| Status         | Draft |
+| Authors        | [Lindani Phiri](@lphiri) |
+| Supersedes     | N/A |
+| Superseded by: | N/A |
+| Tickets        | [RHAIRFE-2922](https://redhat.atlassian.net/browse/RHAIRFE-2922), [RHAISTRAT-2554](https://redhat.atlassian.net/browse/RHAISTRAT-2554) |
+| Other docs:    | operator/multi-tenancy-strategy.md; Tenancy Control Plane design (multitenant-gpuaas: PLAN.md, API-REFERENCE.md, KUEUE-API-REFERENCE.md, ADRs 0001-0033); [RHAIRFE-3115](https://redhat.atlassian.net/browse/RHAIRFE-3115) |
```
davidebianchi: Are those existing files?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
davidebianchi: Since the tenant role is immutable, I agree having different CRDs could simplify

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,1080 @@
...
+  name: demo-team
+spec:
+  quota:
+    resources:
+      - name: gpu
+        type: nvidia.com/gpu
+        guaranteed: "4"
+```
+
+**`TenantProfile`** (auto-created restrictive, then self-managed by the tenant
```
davidebianchi, line 419: Should we set an example for the default auto created `TenantProfile`?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,1080 @@
...
+
+## Open Questions
+
+* **Packaging: embedded platform service vs standalone operator.** The strategy
+  positioned the controller as a platform service inside rhods-operator
+  (alongside Auth, Gateway, Monitoring); the concrete design ships it as a
+  general-purpose operator that also runs on vanilla Kubernetes. Decide whether
+  RHOAI embeds it as a platform service or manages it as a distinct operator
+  enabled via the `Tenancy` component. This affects the operator lifecycle and
+  the vanilla-K8s story.
```
davidebianchi: For the packaging, I think it should be a module. Could be or a standalone module or maybe also part of the cloud controller manager since it setup the infrastructure environment

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,1080 @@
...
+* **No implicit capture of Kueue resources.** Existing ClusterQueues, Cohorts,
+  and LocalQueues are left alone unless explicitly adopted. Generated resources
+  carry `tenancy.opendatahub.io/managed-by: tenancy-controller`; the controller
+  only reconciles and protects resources it owns.
+
+### Onboarding existing namespaces
+
+Adoption is always initiated by the admin, never automatic:
+
+* **Namespace adoption.** Creating a `TenantProject` whose name matches an
```
davidebianchi, line 876: Any namespace can be adopted? Should the PlatformTenant also specify allowed namespace prefix?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,1080 @@
...
+  requires RHBoK 1.4+ (Cohort v1beta2), and there is no OLM dependency to
+  guarantee its presence or version. *Mitigation:* the out-of-band provisioning
+  mechanism installs a compatible RHBoK before the `Tenancy` component is
+  enabled; a version guard surfaces `KueueVersionUnsupported` if an
+  incompatible or missing Kueue is detected at reconcile time.
+* **CRD schema lock-in.** The three CRDs are the root dependency for the tenancy
+  roadmap. *Mitigation:* v1alpha1 with explicit alpha guarantees; design review
+  with downstream owners before v1beta1; per-resource-type quota keyed by Kueue
+  resource names to avoid coupling.
+* **Adoption hard-reconcile can evict workloads.** Adopting an existing CQ
```
davidebianchi: Should we add a webhook on the label addition, to only allow adoption with a matching spec?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+## How
+
+### Three custom resources
+
+All three CRs are cluster-scoped. `PlatformTenant` and `TenantProfile` form a
+1:1 pair with separate authorization domains; `TenantProject` represents a
+namespace inside a tenant.
+
+`PlatformTenant` is named `PlatformTenant` (not `Tenant`) deliberately, to avoid
+collision with the MaaS `maas.opendatahub.io Tenant`/`AITenant` CRDs.
```
lburgazzoli, line 138: The conflicts would only impact `kubectl` , fot GitOps and other clients the problem do not exist since the types belongs to different groups. 

Not a show stopper, but given changing an API name is not easy, I would recommend to find the best name that represent the actual API scope and to me `PlatformTenant`  sounds more as a technical thing than an organizational element.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+escalate by editing their own hierarchy node.
+
+### Network isolation
+
+Per-`TenantProject` `networkIsolation` generates baseline NetworkPolicies:
+
+* `none` — no policies created.
+* `tenant` — blocks cross-tenant in-cluster lateral movement; allows external
+  egress (S3, registries, cloud APIs) and same-tenant traffic.
+* `strict` — full lockdown; only system namespaces and DNS allowed.
```
lburgazzoli, line 217: what happen to existing network policies or network policies created by other tools ? does the `TenantProject` become the owner of the namespace and prevent conflicting policies ?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+
+System namespaces are auto-discovered from DSCI, Gateway/Istio, and Kuadrant and
+labeled for selector-based policies. Component-created NetworkPolicies layer
+additively; the controller never deletes policies it does not own. Cross-tenant
```
lburgazzoli, line 221: Isn't this conflicting with the `strict` network isolation ?  someone creating a `TenantProject` with `strict` policy would genuinely assume the namespace is fully locked, however if another tool create an additional network policy, then it could break the isolation

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+apiVersion: tenancy.opendatahub.io/v1alpha1
+kind: TenantProfile
+metadata:
+  name: research-division       # the root PlatformTenant
+spec:
+  admins:
+    - kind: Group
+      name: research-admins
+  observability:
+    dashboards: true            # tier 2: Perses dashboards over platform Prometheus
```
lburgazzoli, line 451: nit: we should avoid booleans

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+  name: nlp-team
+spec:
+  admins:
+    - kind: Group
+      name: nlp-team-leads
+  defaults:
+    networkIsolation: tenant
+    maxProjects: 10
+  services:
+    maas:
```
lburgazzoli, line 494: using this approach would eventually make the CRD grow very quickly and then causing troubles when services may require an api change or what not. 

Instead, should we use a "binding" approach? 

```yaml
apiVersion: tenancy.opendatahub.io/v1alpha1
kind: TenantProfile
metadata:
  name: nlp-team
policies:
  - apiVersion:  tenancy.opendatahub.io/v1
    kind: MaaSPolicy
    name: foo    
  - apiVersion:  tenancy.opendatahub.io/v1
    kind: KueuePolicy
    name: bar    
```

This would make the Tenant Profile extensible.
Cons: more types and objects but for a GitOps set up it shouod not be a problem

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+  tenant: nlp-team              # owning PlatformTenant; immutable
+  users:
+    - kind: Group
+      name: nlp-team
+      role: edit                # edit | view only
+  networkIsolation: tenant
+  networkGrants:
+    - to: nlp-team/shared-data  # <tenant>/<project>; rules created on both sides
+      direction: egress
+      ports: [8080]
```
lburgazzoli, line 475: should this be like:

```yaml
network:
  isolation: tenant
  grants:
  - ..
```

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -141,81 +137,89 @@ namespace inside a tenant.
...
-available on `TenantProfile.spec.overrides`:
-
-* `hard` — fully isolated; no borrowing or lending.
-* `guaranteed` — donates unused quota to siblings but never borrows.
-* `balanced` — full sharing within the Cohort (Kueue default).
-* `burst` — hoards its own quota but can borrow up to a cap for peak demand.
+The split exists for per-resource RBAC. `PlatformTenant` records a tenant's
+existence and its position in the tree (`spec.parent`); only parent admins (or
+cluster-admin for roots) may create or delete it, so a tenant admin cannot
+re-parent or fabricate tenants. `TenantProfile` is auto-created 1:1 with
```
davidebianchi, line 143: > so a tenant admin cannot re-parent or fabricate tenants

A tenant admin cannot create a new tenant inside the specific tenant?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -141,81 +137,89 @@ namespace inside a tenant.
...
+  over, but only by first rewriting `spec.admins`, which is an explicit, audited
+  act rather than silent ambient access.
+* **Namespace and data access - never inherited.** Access to a project's
+  namespace comes solely from `TenantProject.spec.users`; ancestor admins get no
+  automatic `kubectl` or data access to descendant workloads.
+
+Delegation is protected by three layers:
+
+1. **RBAC** grants tenant admins CRUD on `TenantProfile`/`TenantProject` and
+   create-only on child `PlatformTenant`.
```
davidebianchi, line 192: Should be also able to change/delete a child PlatformTenant (based on the validation webhook checks)?

architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
...
+| Cost accountability | Mandatory ownership metadata, usage metering, allocation rules, showback and chargeback reports | Teams understand what they consume and can manage their spend |
+| Business-level economics | Cost-to-serve by tenant, product, environment, or customer | Finance and product leaders can assess margin, pricing, and investment decisions |
+
+Grouping tenant objects, restricting tenant access, controlling resource consumption, isolating network traffic, and enforcing baseline deployment standards are core multi-tenancy capabilities. Cost allocation should then assign shared CPU, memory, storage, network, and other infrastructure costs proportionally to the tenants that consume them.
+
+## Design principles
+
+- **Start with the tenant experience.** Define what a tenant receives: access model, deployment path, service levels, observability, support, quotas, and cost reports.
+- **Offer tiered tenancy.** Provide shared namespace, virtual cluster, dedicated cluster, or dedicated infrastructure options according to risk, regulatory needs, performance requirements, and willingness to pay.
```
andyatmiami, line 25: While I can appreciate this as a design principal - it seems to be a pretty big ask in light of the [`### Tenant lifecycle` section](https://github.com/opendatahub-io/architecture-decision-records/pull/156/changes#diff-de24e5ef2d1f5e3f05d639142cabdd0f9425401fe98edd19080afcd6aa0ab828R34)

I guess in fullness of time this could theoretically be achievable - but is the thought really that if "dedicated cluster" or "dedicated infrastructure" option is selected - our platform (presumably rhods-operator?) would be tasked with orchestrating that "spin up" ?

architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Resource fairness
+
+Quotas and limits establish the shared-resource contract. They protect cluster stability, prevent accidental overconsumption, and help the platform forecast capacity, while priority and escalation paths distinguish critical production workloads from lower-priority development and batch usage.
+
+Resource controls also support cost governance because requests and usage are necessary inputs when allocating node-level costs to pods.
+
+### Cost attribution
+
+Cost attribution must be a first-class framework capability, designed alongside tenant identity rather than added after workloads are running. Kubernetes spend is commonly billed at underlying infrastructure levels, so the platform must allocate node and shared-service costs back to workloads and tenants using usage, requests, and governed metadata.[1]
```
andyatmiami, line 60: Apologies if I am blind - but I don't see where this `[1]` footnote is then subsequently defined?

architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
...
+
+At minimum, the framework should support allocation by:
```
andyatmiami, line 62: IIUC, MaaS now has a token/usage driven allocation... and its unclear if/how that should relate to the framework allocation support listed here... as there runs a real (obscure/difficult to quantify) risk of "double-counting" here, right?

architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
...
+| Security | Isolation standards, policy guardrails, risk tiers, audit and exception governance |
+| FinOps / finance | Allocation model, financial-data reconciliation, showback/chargeback, budgets and reporting |
+| Application teams | Accurate metadata, efficient resource configuration, remediation of idle or overprovisioned workloads |
+| Product / business owners | Tenant profitability, service-tier choices, pricing and investment decisions |
+
+FinOps practices for Kubernetes commonly include mapping costs to applications and business units, then using the resulting data for budgets, forecasts, dashboards, and scorecards.
+
+## Measures of success
+
+Measure customer and business outcomes alongside control coverage:
```
andyatmiami, line 102: Do we have means of actually collecting this data? Is the thought we'd include "phone home" logic that periodically polls tenant-related operations on a customer's cluster to evaluate these metrics and send them back to us "anonymized" ?  I wholly agree with the metrics.. its just not clear (to me) how **WE** measure them...

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+  within their subtree without platform-admin intervention.
+* Provide network isolation presets (none/tenant/strict) with cross-tenant
+  grants.
+* Provide tiered observability: always-on metric labeling, opt-in Perses
+  dashboards, and opt-in per-tenant dedicated monitoring for showback/chargeback.
+* Provide delegated MaaS tenant provisioning built on the tenant identity.
+* Activate capabilities by CRD detection so the control plane degrades
+  gracefully and runs on both OpenShift and vanilla Kubernetes.
+* Preserve full backward compatibility: unmanaged namespaces operate unchanged;
+  adoption of existing namespaces is opt-in.
```
andyatmiami, line 100: In these hybrid scenarios... how does that play into the cost attribution substrate... some thoughts
- each existing namespace is treated as its own "tenant"
- all existing (unmanaged) namespaces are rolled into a single "nameless tenant"
- ignored from cost attribution substrate

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+
+## Non-Goals
+
+* Resource quota and fairness. GPU/compute quota, borrowing/preemption, and
+  Kubernetes ResourceQuota are out of scope for this ADR and deferred to a
+  follow-up (for example a Kueue-based allocation model layered on this tenant
+  identity). This framework provides the tenant hierarchy those systems attach
+  to; it does not allocate or enforce resources.
+* Multi-cluster tenancy federation. Single-cluster scope; cross-cluster is
```
andyatmiami, line 109: This working confuses me... is "Single-cluster scope" a non-goal?  or we are saying Multi-cluster tenancy federation is a non-goal - with then the "qualifier" that we are focusing on Single-cluster scope?

If the latter - seems like that should be explicitly called out in `## Goals` section?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+
+`spec.parent` is immutable. An operator-level setting may impose a conservative
+maximum depth as a safety guard, but the framework does not promote deep trees.
+
+### Delegated administration and authorization
+
+Administration is delegated down the tree. A parent admin creates a child
+`PlatformTenant` and bootstraps its initial `TenantProfile.spec.admins`; from
+then on those tenant admins self-manage the profile and create projects and
+sub-tenants within their subtree, up to `TenantProfile.spec.defaults.maxProjects`.
```
andyatmiami, line 165: `TenantProfile.spec.admins` : can this be users, groups, or both?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+  orphan-recovery path: if a tenant's admins all leave, an ancestor (ultimately
+  cluster-admin) reassigns them. Because this authority is computed from the tree,
+  not from being listed in the child's `spec.admins`, a tenant admin cannot lock
+  an ancestor out.
+* **Configuration self-management - not inherited.** Only a tenant's own
+  `spec.admins` may edit the rest of its `TenantProfile` (network defaults,
+  observability, `services`) and its `TenantProject`s. An ancestor can still take
+  over, but only by first rewriting `spec.admins`, which is an explicit, audited
+  act rather than silent ambient access.
+* **Namespace and data access - never inherited.** Access to a project's
```
andyatmiami, line 185: How does observability data fit into this?  I appreciate the existing rationale here.. but i feel prometheus/loki data maybe warrants its own call out here... as its not clear if "data access to descendant workloads" includes (or excludes) this...

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+  namespace comes solely from `TenantProject.spec.users`; ancestor admins get no
+  automatic `kubectl` or data access to descendant workloads.
+
+Delegation is protected by three layers:
+
+1. **RBAC** grants tenant admins CRUD on `TenantProfile`/`TenantProject` and
+   create-only on child `PlatformTenant`.
```
andyatmiami, line 192: child `TenantProfile` / `TenantProject` aren't even _view-only_ ( `get / list / watch` )? for parent tenant admins?

no problem here regardless.. it just isn't explicitly stated so I wanted to raise for clarity..

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+These are filtered views over shared data, not isolation. They surface generic
+namespace and workload metrics (`container_cpu_usage_seconds_total`,
+`container_memory_working_set_bytes`, pod/workload counts, and any
+component-exposed application metrics), giving tenants self-service visibility
+without the cost of a dedicated stack.
+
+**Tier 3 - dedicated monitoring (opt-in, root-only).** When a root
+`TenantProfile` sets `observability.dedicatedMonitoring.enabled: true` and the COO
+CRDs are present, the controller provisions a dedicated monitoring namespace
+`<prefix><root>-monitoring` containing:
```
andyatmiami, line 268: So does Tier 3 end up potentially impacting the "cluster admin" persona? If some **super admin** wanted to query <whatever> across the cluster as a whole... are we providing tools, etc to facilitate that?

andyatmiami: Ah, I see this is specifically referenced/addressed here:
- https://github.com/opendatahub-io/architecture-decision-records/pull/156/changes#diff-18d1f618cbf211a6c73e07abfb2b02fcbe87fb28ea282811c7c014d622722000R309-R312

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+* Lifecycle follows ownership: enabling creates the `AITenant`; disabling it, or
+  deleting the tenant, deletes the `AITenant` and triggers maas-controller's
+  existing cascade cleanup (api-key revocation, maas-api teardown, namespace and
+  policy deletion). The Gateway CR stays network-admin-owned; if it is missing,
+  the `AITenant` condition surfaces the error up to the tenant status, preserving
+  the MS-0003 coordination constraint.
+* Backward compatibility: hand-created `AITenant`s (including the default
+  `models-as-a-service` tenant) keep working untouched; the controller manages
+  only `AITenant`s it generated. An existing `AITenant` can optionally be adopted
+  by binding it to a tenant.
```
andyatmiami, line 382: Same question w.r.t. cost attribution substrate as here:
- https://github.com/opendatahub-io/architecture-decision-records/pull/156/changes#r3866494397

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+apiVersion: tenancy.opendatahub.io/v1alpha1
+kind: TenantProject
+metadata:
+  name: sentiment-analysis
+spec:
+  tenant: nlp-team              # owning PlatformTenant; immutable
+  users:
+    - kind: Group
+      name: nlp-team
+      role: edit                # edit | view only
```
andyatmiami, line 470: Is this `users` block **required** ? I'm curious how this intersects with some of the _granular RBAC work_ that has (slowly) been chugging along in RHOAI...

I imagine this clearly needs to still tolerate native k8s RBAC assignment - so is this `users` block simply a convenient mechanism to seed initial users?

Would like more information on what `edit` + `view` _imply here_ ? Are there references to native k8s aggregated roles?  or just general conceptual pointers?

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+        certificateRef:
+          name: nlp-team-maas-tls
+          namespace: ai-tenants
+      quotas:                        # MaaS control-plane counts
+        maxModels: 20
+        maxSubscriptions: 100
+        maxApiKeys: 50
+```
+
+**Generated `AITenant`** (rendered and owned by the tenancy controller; the
```
andyatmiami, line 511: So I guess this creates a coupling between the `TenantProfile` CRD and `AITenant` CRD?

If `AITenant` is enhanced - that **requires** accompanying changes to `TenantProfile`?   I can appreciate wanting to keep them separate to support standalone MaaS capabilities.. but this "CRD shadowing" is probably a risk worth explicitly calling out? If I am understanding this correctly..

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: Completely open to other names here. Naming is hard and I would welcome suggestions here.
Organization/Team/Account/Project    were all alternatives I had considered here.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+  carrying the `tenancy.opendatahub.io/tenant` label. A namespace without that
+  label is never touched.
+
+### Onboarding existing namespaces
+
+Adoption is always initiated by the admin, never automatic:
+
+* **Namespace adoption.** Creating a `TenantProject` whose name matches an
+  existing namespace adopts it instead of failing: the controller adds tenant
+  labels and applies the requested NetworkPolicies and RBAC. The `NamespaceReady`
```
andyatmiami, line 878: Existing NetworkPolicies and RBAC are, I am assuming, left in place? Which then means the manifest we apply may actually be undercut?  Granted, that is "how k8s works" - but should we do any type of warning? Or at minimum documentation?  I don't think we should **solve** this problem - but I think we need to someone call out the "risks"..

tl;dr I worry adopted namespaces could in practice have loose security than "from-creation managed" namespaces.. and that could "suprise" someone..

andyatmiami (commented):
Hey folks.. first time I have really weighed in on an ADR - so "please be nice to me.. unless I deserve it"

Great write up thus far - I have some questions posted here that generally revolve around the theme of making sure I appreciate and understand the system being built.

Its possible (likely?) _answering these question_ may not be "in scope" to the ADR - as at least a few times I certainly wade into "implementation details" discussions..

But, look forward to see this proposal evolve and become reality - and happy to help if/as I can to reach that state!

architecture-decision-records/operator/multi-tenancy-strategy.md
lphiri: @andyatmiami  this is high level strategic view - its not intended that everything will be implemented by our platform.  Dedicated cluster option would be  be done by a  high level orchestrator  (ACM or HCP and the like).  rhods-operator would only deal with incluster level  "soft tenancy".

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: Ack, will rephrase to make things clearer.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: Good catch, a tenant can create subtenants (with constraints)

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: can be both

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: yes

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: Yes, this is a challenge with NP in general.    we will iterate on this portion once the current NP audit and hardenign work is complete.
The general idea though is to start with a secure default.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: We will need to iron out the details of what it means to adopt an existing project.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: Ack - will keep this open to refactoring as the POC progresses.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: Yes this is a risk with any abstration in general.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: Good question.   My thinking is that adoption would be done by the tenant admin creating a tenantproject , and not simply labelling a namespace.   However, we need to work through the implications in a little more detail in the POC

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
lphiri: See comment above from Luca -  we need to work through this one 
https://github.com/opendatahub-io/architecture-decision-records/pull/156/changes#r3860772264

andyatmiami: will use luca's comment to continue any discussion and resolve this thread - thanks for pointer!

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
davidebianchi: Should we maybe setup our network policy, detect already existent ones and raise in the status? In this way, the user is aware that exists some extraneous network policies which could conflict with the strict network isolation

bryonbaker (commented):
This is a very good description but I think it centres on the tenant user experience too much. This needs to be a balanced description of the needs of a tenant and the service provider.  In fact you could argue that the tenant experience is nothing more than a native Kubernetes and OCPAI experience and they are unaware that they are on a shared platform. That is the core requirement. There are then a multitude of requirements that immediately fall out of that core requirement.

I think the tenant's requirements are:

Each tenant should experience the service as an independent Kubernetes environment:

* Native Kubernetes APIs and tooling.
* Their own logical Kubernetes identity and administrative boundary.
* Their own namespaces, workloads, policies, quotas and configuration.
* Their own endpoint, branding and service identity where appropriate.
* No requirement to understand that infrastructure is shared.
* No visibility into other tenants or provider implementation details.
* Behaviour sufficiently Kubernetes-native that existing Kubernetes tooling and deployment practices continue to work.

A more balanced perspective could be:
| Tenant perspective | Service-provider perspective |
| --- | --- |
| Fast onboarding | Low-cost automated provisioning |
| Individual user experience | White labelling |
| Isolation | Controlled blast radius |
| Predictable performance | Capacity efficiency and service stability |
| Self-service | Reduced operational toil |
| Observability | Fleet-wide operational visibility |
| Explainable costs | Cost recovery and margin visibility |
| Choice of tenancy tiers | Standardised, supportable service catalogue |
| Clear service levels | Measurable SLOs and operational ownership |
| Easy lifecycle | Repeatable upgrades, migrations and retirement |

architecture-decision-records/operator/multi-tenancy-strategy.md
```diff
@@ -0,0 +1,146 @@
...
+- **Make costs explainable.** A chargeback number without a transparent allocation method creates disputes; show the source usage, shared-cost logic, and allocation dimensions.
+- **Separate visibility from enforcement.** Start with showback so teams can understand costs, then introduce budgets, alerts, optimization targets, and chargeback where the organization is ready.
+
+## Framework pillars
+
+### Tenant lifecycle
+
+The platform should automate tenant creation, entitlement changes, and retirement. A tenant blueprint can create the namespace or virtual cluster, RBAC bindings, default network policies, quota objects, metadata schema, monitoring integration, and cost-reporting identity in one workflow.
+
+This reduces management effort for the platform team and creates a repeatable customer experience. It also eliminates configuration drift that arises when tenancy is provisioned through tickets and manual runbooks.
```
StevenTobin, line 38: With modularity the platform has much less control over the resources that are created on the cluster. We should probably plan to create a new MT module that manages this rather than rolling it into the platform operator itself. 

We would also need an agreed contract between the other modules and the MT module defining what they need to do with each of these resources. The expectations for what it means when the MT module creates a virtual cluster or tenant CR would need to be well understood and tested for each module individually. We could codify these requirements into tests and functions in the [odh-platform-utilities ](https://github.com/opendatahub-io/odh-platform-utilities) library.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+| Tickets        | [RHAIRFE-2922](https://redhat.atlassian.net/browse/RHAIRFE-2922), [RHAISTRAT-2554](https://redhat.atlassian.net/browse/RHAISTRAT-2554) |
+| Other docs:    | operator/multi-tenancy-strategy.md; Tenancy Control Plane design (multitenant-gpuaas: PLAN.md, API-REFERENCE.md); [RHAIRFE-3115](https://redhat.atlassian.net/browse/RHAIRFE-3115) |
+
+## What
+
+RHOAI will introduce a top-level multi-tenancy framework: a first-class tenancy
+control plane that represents an organizational tenant hierarchy and provides
+governed, self-service namespace provisioning with delegated administration,
+network isolation, per-tenant observability, and delegated Models-as-a-Service
+(MaaS) provisioning. The framework is delivered as a controller plus a
```
StevenTobin, line 20: I think this should be a new multitenancy module that we integrate with the platform operator and not functionality in the platform operator itself.

architecture-decision-records/operator/multi-tenancy-strategy.md
davidebianchi: Yes, the idea is that it is a standalone tenancy module

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+* **Cost attribution:** always-on metric labeling plus optional per-tenant
+  dashboards and dedicated monitoring provide the consumption substrate for
+  showback and chargeback.
+
+The strategy's **resource fairness** pillar (quota, limits, priority) is
+acknowledged but deferred to a follow-up ADR so this foundational tenant model
+can land first. Without a structured tenant model, RHOAI cannot serve as a
+shared platform at enterprise scale and loses ground to AI platforms that ship
+multi-tenant governance out of the box.
+
```
sefroberg, line 80: I think a subsection defining the users, roles or persona is needed to make sure we account for the different levels of admin and users that could be involved with this functionality.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
sefroberg: My understanding is a tenant can 0...many namespaces each.

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+| Scope          | Operator (RHOAI platform-wide multi-tenancy) |
+| Status         | Draft |
+| Authors        | [Lindani Phiri](@lphiri) ,  [Chris Sams](@csams)|
+| Supersedes     | N/A |
+| Superseded by: | N/A |
+| Tickets        | [RHAIRFE-2922](https://redhat.atlassian.net/browse/RHAIRFE-2922), [RHAISTRAT-2554](https://redhat.atlassian.net/browse/RHAISTRAT-2554) |
+| Other docs:    | operator/multi-tenancy-strategy.md; Tenancy Control Plane design (multitenant-gpuaas: PLAN.md, API-REFERENCE.md); [RHAIRFE-3115](https://redhat.atlassian.net/browse/RHAIRFE-3115) |
+
+## What
+
```
sefroberg, line 15: Another point. I think we need to clearly define what we are talking about for multi-tenancy. The document still has some points where it blends Soft Multi-tenancy and Hard Multi-Tenancy. You sprinkle your ideas throughout the doc but clearly defining, "This is what our plan for multi-tenancy for OCP/K8s and RHOAI is this."

architecture-decision-records/operator/ODH-ADR-Operator-0015-multi-tenancy-framework.md
```diff
@@ -0,0 +1,989 @@
...
+* Hardware/accelerator access gating (DeviceClass, StorageClass). Deferred, and
+  naturally paired with the future quota ADR.
+* Automated hierarchy re-parenting. A tenant's parent is immutable; reorganizing
+  is done by creating new nodes and migrating.
+
+## How
+
+### Three custom resources
+
+All three CRs are cluster-scoped. `PlatformTenant` and `TenantProfile` form a
```
sefroberg, line 133: I think we need to look at the ability for the CR's to be either cluster-scoped or Namespaced. And maybe some rule that a Namespaced CR cannot be a root in the tree. The problem I am concerned about is a cluster-scoped CR.

davidebianchi: Which problem are you thinking about for cluster-scoped CR in this case? Since those are related to namespaces/projects and above hierarchy, not sure those could be Namespaced.
e.g. the TenantProject one, which is the last in the hierarchy, should trigger the creation of a ns/project so it should be cluster scoped to own resources in that namespace (since it should be created before the ns itself).

PR #157, opened 2026-08-24
doc: Per-model rate limits

mariusdanciu:
## Description
- ADR for allowing specification of per-model rate limits irrespective of the users and MaaS subscriptions.

## Merge criteria:

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record for optional per-model token quota limits.
  * Documented quota interactions with subscription limits, lifecycle management, status reporting, enforcement behavior, and rate-limit responses.
  * Added an architecture decision record describing bring-your-own-connector support for Data Connect Hub.
  * Documented custom connector packaging, deployment, routing, registration, security considerations, risks, alternatives, and open questions.
  * Recorded alternatives considered for model-level quota enforcement.


architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
```diff
@@ -0,0 +1,311 @@
+# Open Data Hub - Per-Model Quota Limits
+
+|                |            |
+| -------------- | ---------- |
+| Date           | 2026-08-24 |
+| Scope          | AI Gateway, Models-as-a-Service (MaaS), Quota Management |
```
shaneutt, line 6: In AI Gateway @jland-redhat and I already have a design in Praxis for tokenomics, and we have epics and a project board there for TRL that is underway. Notably, we will be supporting TRL in other environments (including non-maas, and even non-k8s), so it doesn't appear to me that investing in expanding MaaS itself with per-model quotas aligns logistically or architecturally. I would like to get Jamie's thoughts on that (please) 🤔 

We will need to coordinate with Jamie, and then upstream Praxis. Let me know your thoughts? 🙇

shaneutt (dismissed):

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
mariusdanciu: If this will be implemented in Praxis without Limitador yes this is a very different path and this ADR won't apply. So I have a few questions:
1. When is this targeted ? 3.7 ?
2. Is there a separate ADR or design doc on how this will work in Praxis ? (Control plane and data plane ?)
3. These are significant architectural changes that impact MaaS that I need to be included/aware. Discussing in different streams is not ideal. I am aware that eventually we want to do these in Praxis and move away from Limitador (definitely agree) but I wasn't aware about this existent effort. 

BTW there are other discussion around per org limits that does not involve praxis at all. Please see https://redhat-internal.slack.com/archives/C0BN21A8J68/p1786350727777959 ... just to add more confusion into the mix. So by the above reasoning this would also not architecturally align. Unless we do these for Limitador for now (tactically) and move them to Praxis later on ?

jland-redhat: Might need to re-sync on this, but I know y'all's tokenomics is essentially just a different implementation of Limitador's TRLP-though still very non-opinionated like Kuadrant is today.

That means MaaS will play the same role it did before. It'll live on top as the opinionated way to define those limits using MaaS Models/Subscriptions/AuthPolicies to create the underlying policies.

If the main worry is generating Kuadrant TRLP rather than Praxis's new Kubernetes object, I wouldn't put too much stock in that. The translation to TRLP is simple and won't look much different from our future setup anyway. The real challenge is just defining the CRD itself (which we'll need to do regardless of any subsequent feature changes).

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
```diff
@@ -0,0 +1,311 @@
...
+    EndpointOverride string `json:"endpointOverride,omitempty"`
+    // TenantRef is the name of the AITenant this model belongs to
+    // +optional
+    TenantRef string `json:"tenantRef,omitempty"`
+    // QuotaLimits defines model-wide token rate limits that cap aggregate
+    // consumption across all users and subscriptions. These limits protect
+    // the inference backend regardless of how many subscriptions reference
+    // this model.
+    // +optional
+    QuotaLimits []TokenRateLimit `json:"quotaLimits,omitempty"`
```
jland-redhat, line 63: I still don't personally think this makes sense. I know this was something BofA mentioned through Javo but do we have any other examples of folks asking for this?

Also adding quota logic on the model itself does increase our general complexity of the rules we need to create.

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
```diff
@@ -0,0 +1,311 @@
...
+    │       YES ↓
+    │
+    ├── Model Quota TRLP: is the model within its global limit?
+    │       NO  → 429 (model quota exceeded)
+    │       YES ↓
+    │
+    └── Request forwarded to inference backend
+```
+
+Kuadrant evaluates all TRLPs targeting the same HTTPRoute. Both must pass. The response headers should distinguish which limit was hit (Kuadrant includes `X-RateLimit-*` headers from Limitador).
```
jland-redhat, line 191: I don't know this is part of Kuadrant yet but I also don't know it is relevant for what we are trying to deliver here.

And may be different if we swap out using Praxis

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
```diff
@@ -0,0 +1,311 @@
...
+
+### Shared HTTPRoute Consideration
+
+The existing limitation documented in [quota-and-access-configuration.md](docs/content/configuration-and-management/quota-and-access-configuration.md) applies here as well: when multiple MaaSModelRefs share the same HTTPRoute, their quota TRLPs may conflict. The same mitigation applies — prefer dedicated routes per model. Because model quota TRLPs are owned by the MaaSModelRef (not a subscription), the conflict surface is simpler: at most one quota TRLP per MaaSModelRef per HTTPRoute.
+
+## Open Questions
+
+1. **Metrics and observability**: Should the controller expose Prometheus metrics for model quota utilization (e.g., `maas_model_quota_usage_tokens_total`, `maas_model_quota_remaining_tokens`)? Limitador already exposes counters, but a MaaS-level metric may be easier to consume.
```
jland-redhat: So "remaining quota" is something that is an upstream ask and would not be something that you can expose through promtheus since it is only relevant at a single point in time.

`maas_model_quota_usage_tokens_total` can already be inferred through the `authorized_hits` metric that gets reported since that metric includes the model

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
```diff
@@ -0,0 +1,311 @@
...
+
+2. **ExternalModel support**: ExternalModels may have their own provider-side rate limits. Should the quota defined here apply only to the gateway layer, or should it also account for provider limits? Initial proposal: gateway-only, document that ExternalModel providers have their own limits.
```
jland-redhat: From a Token Rate Limiting perspective we don't distinguish between Internal and External models. We do it at the MaaSModelRef level so anything that applies to an Internal model would also apply to an External Model

jland-redhat (commented):
To be honest I still don't know I see the value of this and it does add complexity.

I think having a limit for a model at the Subscription (or Subscription Metadata) level is more valuable an a much easier task to implement.

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
mariusdanciu: I agree. I see a short term path using the existent TRLP and later on move things to Praxis. IMO this should be done after we are full on Praxis, no praxis-ext_proc, no envoy.

But based on the slack conversation we need to clarify the priority of per-org level rate limits and per-model. If there are customers asking explicitly for per-org rate limiting I am OK to prioritize this first, but the ADR should address both. @jland-redhat any thoughts ?

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
mariusdanciu: That's fine. But imo user level and org level quotas do not protect the model itself. No one has infinite resources to scale up llms so IMO we should definitely provide clients a way to protect the model itself.

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
mariusdanciu: This is what the statement it. If this is not needed explicitly I'm ok with it.

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
mariusdanciu: Agreed although the sentence raises the point that the external provider of the model might have other quotas already configured that we don't know about. But yes a bit of rephrasing makes sense.

architecture-decision-records/model-serving/ODH-ADR-MS-0005-per-model-quota-limits.md
shaneutt: After talking with Jamie we're going to confirm some things with the PMs, but I'm removing my block based on the understanding that we _do_ in fact want MaaS to be used as a higher-order API for this kind of thing, though there will still be the lower-order (and independent) parts.

PR #159, opened 2026-08-31
ODH-ADR-MX-0001: ModelExpress metadata service deployment model

wseaton:
Adds the first ADR for the ModelExpress component, covering how the metadata service deploys on a cluster. The default topology is a shared singleton in a system namespace, with namespace-scoped instances as an opt-in for tenants that need metadata isolation. ServiceAccount token auth is the access boundary.

For this release, the KServe operator deploys and manages the ModelExpress operator. Serving is the only consumer, and KServe already templates `LLMInferenceService` workloads against the metadata service. ModelExpress keeps its own APIs and reconciler, so ownership can move out of KServe once a non-serving consumer such as RL lands.

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record describing the ModelExpress metadata-service deployment model.
  * Documented shared-cluster and namespace-scoped deployment options, authentication, and lifecycle management through the AI Gateway operator.
  * Documented integration with KServe, reinforcement learning workloads, and local model caching.
  * Added deployment alternatives, security considerations, operational risks, stakeholder impacts, and implementation references.


danielezonca:
@mariusdanciu @pierDipi 
can you please have a look?

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
```diff
@@ -0,0 +1,100 @@
...
+| Status         | Draft |
+| Authors        | [Will Eaton](@wseaton) |
+| Supersedes     | N/A |
+| Superseded by: | N/A |
+| Tickets        | |
+| Other docs:    | [ModelExpress upstream](https://github.com/ai-dynamo/modelexpress), [opendatahub-io/modelexpress](https://github.com/opendatahub-io/modelexpress) |
+
+## What
+
+How the ModelExpress metadata service is deployed on a cluster: how many instances run, where, and in which namespace its metadata CRs (`ModelMetadata`, `ModelCacheEntry`) live. The decision: a shared cluster singleton in a system namespace by default, with namespace-scoped instances as an opt-in isolation flow.
```
mariusdanciu, line 16: ModelMetadata CR name can seem a bit generic. Can you please provide examples of these CRs ?

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Why
+
+ModelExpress deduplicates model downloads and coordinates peer-to-peer weight transfer between inference workloads. Its metadata service tracks which models exist on the cluster and where, using namespaced CRs as its state store. The benefit grows with the number of workloads sharing one metadata view: two namespaces serving the same model only deduplicate if they share a metadata service. That argues for one shared instance. Some tenants, however, must not expose even model names across namespace boundaries, so an isolated topology has to exist too.
+
+ModelExpress is enabled through the `DataScienceCluster` CR, and this ADR proposes it as a root-level component: a peer of Model Serving, not a sub-component of it, with no runtime dependency on KServe or any other component operator.
```
mariusdanciu, line 28: I am not sure MX should be a top level DSC component. It cannot exist without KServe and ODH does not have a component dependency mechanism. Why not integrate these CRs in KServe controller or even integrate this MX metadata in LLMISVC ?

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Non-Goals
+
+* The metadata CR schemas themselves; those are owned upstream.
+* Hard singleton enforcement; the default is a convention, not a CEL constraint.
+* The `DataScienceCluster` component API surface for ModelExpress (fields, defaults); that follows the standard component onboarding process.
+
+## How
+
+The metadata service is deployed via a namespaced `ModelExpressServer` CR reconciled by the ModelExpress operator, using upstream's Kubernetes CRD metadata backend. All ModelExpress CRDs are namespace-scoped.
```
mariusdanciu, line 45: Do we need a separate MX Operator ? Is there a reason why we cannot add MX capabilities in KServe ?

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
pierDipi, line 16: can't we re-use and expand Kserve's LocalModelCache instead of adding another ModelCacheEntry? That could reduce user confusion and give us a path to hardware agnostic API as now ModelCacheEntry is in the modelexpress.nvidia.com group

wseaton: It's not my name. It comes from upstream: https://github.com/ai-dynamo/modelexpress/blob/8cbca13c1ea3bd0576fe0073cbd1e7653044e4af/modelexpress_server/src/p2p/k8s_types.rs#L25

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
wseaton: Again, this ties MX to a serving orchestrator when it is also heavily used in RL workloads. We should keep the control plane orchestration standalone, and KServe should handle workload templating to integrate. KServe does not need to manage deployments of the metadata coordinator.

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
wseaton: Can you elaborate on what this would look like practically?

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
wseaton: There are use-cases for modelexpress outside of pure serving, it is widely used in RL communities to accelerate post-epoch rollout. This would be used without KServe.

wseaton:
Added a KServe integration section under "How"

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
```diff
@@ -0,0 +1,120 @@
...
+
+**Install.** The ODH operator deploys the ModelExpress operator when the component is enabled in the `DataScienceCluster`. The operator watches `ModelExpressServer` CRs in all namespaces and requires no other component to be enabled.
+
+## KServe Integration
+
+The metadata service and KServe operate at different layers. The metadata service is cluster infrastructure: admin-provisioned, auth-configured, consumed by workloads across namespaces and orchestrators. KServe is one consumer. Its responsibility is workload templating, not metadata service lifecycle.
+
+### Metadata service deployment
+
+The ModelExpress operator is a standalone `DataScienceCluster` component. KServe does not install or manage `ModelExpressServer` instances. An admin creates the CR, configures the ServiceAccount allowlist, and the operator publishes the gRPC endpoint on `status.endpoint`. KServe workloads consume that endpoint.
```
spolti, line 61: How will this be consumed, will the vllm runtime that will consume the API or?

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Alternatives
+
+* **Cluster-scoped metadata CRDs with one mandatory instance.** Trivial discovery, but per-tenant visibility becomes impossible without admission-level filtering, and the service needs cluster-scoped write access. The isolation flow stops being implementable.
+* **Namespace-scoped instances only.** Uniform and isolated, but N namespaces serving the same model means N downloads and N cache copies; the deduplication win disappears exactly where it matters (large shared foundation models).
```
spolti, line 78: I guess that is related with @pierDipi's comments, LocalModelCache can still help in this scenario, where the model is still downloaded just once, and instead copied over NS, it is copied over nodes and mounted in the desired workloads.

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
pierDipi: users instead of creating

```yaml
apiVersion: modelexpress.nvidia.com/v1alpha1
kind: ModelCacheEntry
# ...
spec:
  modelName: meta-llama/Llama-3.1-70B
  provider: HuggingFace
```

would create something like

```yaml
apiVersion: serving.kserve.io/v1alpha1
kind: LocalModelNamespaceCache                                  # or cluster-level LocalModelCache
# ...
spec:
  sourceModelUri: hf://meta-llama/Llama-3.1-70B  
  # ... etc
```

This would trigger the model download + metadata storage in the MX server, then llmisvc/model cache controller / webhook injects MX Server coordinates for selected deployments

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
wseaton: Yes, the vllm runtime needs to be configured with env vars that point it to the central `ModelExpressServer` service, as well as enable the `mx` weight loader plugin. From there, vllm can see other peers on startup.

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
wseaton: I added a section on the how that addresses this, I believe it composes. MX uses local disk as a fallback path if no peer holds weights in HBM, this could be a LocalModelCache.

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
wseaton: I think you might be misunderstanding the model, `ModelCacheEntry` is an internal CRD used by modelexpress. It is not meant to be user driven to hydrate a cache, it is for bookkeeping when using the state backend `kubernetes`.

wseaton: The P2P engine is pull-through, it becomes aware of models when an inference server spins up serving them for the first time

pierDipi: I see

architecture-decision-records/modelexpress/ODH-ADR-MX-0001-metadata-service-deployment-model.md
pierDipi: Given that use case, I think it seems preferred to have it top-level while initially documentation will remain focused on serving/inference use cases

PR #160, opened 2026-09-03
Openshell dashboard BFF adr 

d0w:
## Description

Adds the first ADR for Openshell integration into ODH dashboard. This ADR primarily involves the initial planning of the upstream dashboard Go BFF and how downstream can consume the BFF if needed. It describes the primary needs of the upstream BFF and cases of how downstream can reuse, modify, or overwrite implementations without needing to fork and rebase upstream.

## Summary by CodeRabbit

- **Documentation**
  - Added an architecture decision record covering the proposed Openshell Dashboard backend-for-frontend architecture.
  - Documented reusable integration and customization patterns, including service boundaries, routing, authentication, middleware, dependency management, and testing.
  - Clarified downstream integration considerations, release coordination, and approaches for reusing upstream components.
  - Recorded alternative solutions and expected impacts on downstream integrations and future component reuse.


rhuss:
@d0w thanks for the ADR. To review it properly I need some more context, and I think the ADR itself would benefit from a broader introduction and a clearer problem statement.

Specifically:

- **Why is a downstream BFF needed at all?** What downstream-specific requirements prevent using the upstream BFF directly? Concrete examples of what downstream adds or overrides would help.
- **What is the ongoing sync model?** When upstream adds new endpoints or changes interfaces, how does downstream pick those up? Is there CI that validates compatibility, or is it manual version bumping?
- **How does this relate to the existing ODH BFFs?** The gen-ai, model-registry, and agent-ops BFFs all use `julienschmidt/httprouter`. This ADR proposes Chi for the upstream BFF. If `/pkg` is the shared code surface, the different handler signatures will be a friction point. Was this considered?
- **What exists today?** The title says "Migration" but the ADR doesn't describe the current state. Is this replacing an existing BFF, or is it net-new?

Happy to do a full review once these are clarified. The Go patterns in the ADR (struct embedding, interface decoration, constructor DI) are solid, so the technical direction looks right. I just want to understand the constraints that shaped the architecture before evaluating the specific choices.

d0w:
@rhuss Thanks for the first pass. I'll expand on the ADR with those questions in mind. Before that though, is there a mandate/reason for why `julienschmidt/httprouter` is used? The only place I found was in [Gen Ai's BFF](https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/gen-ai/docs/adr/0002-system-architecture.md#backend-for-frontend-container-go-126). 

If there is to be a downstream BFF in the odh-dashboard, my understanding is that it would be contained into its own package and the use of `Chi` shouldn't affect other components. It's also built directly on stdlib (similarly to httprouter) which makes the handler syntax nearly identical. There is some friction to address here (since the upstream BFF already uses Chi) and ideally downstream would also use Chi. They both extend stdlib `net/http` so they are interchangeable to an extent. Though, if there are other concerns such as code convention then that's also understandable and I'll adjust the ADR with that in mind.

Edit: If possible, I'd still like to know about the question above but we are probably going to go with using upstream's BFF for now. The shared code surface will stay though to leave room for movement in the future if we do decide we need a separate BFF for different functionality.

rhuss:
I have to confess that the assessment around `julienschmidt/httprouter` was not fully verified, but my feeling was that the various BFFs in RHAI might want to be aligned (just to reduce the maintenance surface), but this is not a show stopper for me (feel free to ignoire that aspect). My main concern indeed is, to stay as close as possible to upstream to benefit from future improvements and additions.

d0w:
@rhuss Hopefully these comments were addressed in the latest updates but for a concise answer:

> Why is a downstream BFF needed at all? What downstream-specific requirements prevent using the upstream BFF directly? Concrete examples of what downstream adds or overrides would help.

~~We decided to reuse the upstream BFF directly for now by taking the built images from upstream and packaging them with the downstream frontend. Though, if we ever need to add custom logic or other additions, the code patterns described would help facilitate a separate BFF without the need for forking and rebasing.~~

We will need to have a separate downstream BFF to support multi-gateway authentication flows and routing. Upstream does not have this requirement (only 1 gateway).

> What is the ongoing sync model? When upstream adds new endpoints or changes interfaces, how does downstream pick those up? Is there CI that validates compatibility, or is it manual version bumping?

Downstream will pin versions to ensure updates don't break downstream automatically. If we decide to use a separate downstream BFF, the imported modules will inherit all the changes. If we add custom logic, we would need to update that code manually to abide by new interfaces or api contracts. It is manual version bumping for now but there are future plans to add CI that will validate compatibility via contract tests. 
    
> How does this relate to the existing ODH BFFs? The gen-ai, model-registry, and agent-ops BFFs all use julienschmidt/httprouter. This ADR proposes Chi for the upstream BFF. If /pkg is the shared code surface, the different handler signatures will be a friction point. Was this considered?

There will be some friction here but the current upstream BFF uses chi and it is a good, lightweight tool for the job. Downstream will ideally use Chi as well but if not, there are ways to adapt the different handlers. If the friction gets too far, the other components such as service functions can be reused while handlers are completely separate.
    
> What exists today? The title says "Migration" but the ADR doesn't describe the current state. Is this replacing an existing BFF, or is it net-new?

A backend exists [here](https://github.com/Gkrumbach07/openshell-dashboard/tree/main/backend) and this is intended to refactor that. The API behavior before vs. after the refactor should be the same though.

architecture-decision-records/openshell/ODH-ADR-0001-openshell-bff-migration-architecture.md
```diff
@@ -0,0 +1,491 @@
...
+# What
+
+Proposal for changes to the Openshell Dashboard Go BFF in the upstream ([https://github.com/NVIDIA/OpenShell](https://github.com/NVIDIA/OpenShell)) and how downstream ([https://github.com/opendatahub-io/odh-dashboard](https://github.com/opendatahub-io/odh-dashboard)) will reuse and modify upstream if needed.
+
+## Key Points
+
+- Struct embedding and decoration as a code pattern to maintain interoperability between upstream and downstream BFFs  
+- `Chi` ([GitHub \- go-chi/chi: lightweight, idiomatic and composable router for building Go HTTP services](https://github.com/go-chi/chi)) for easier routing and middleware handling  
+- A revamped file structure including the separation of handlers into separate objects to accompany interface decoration
+- Downstream will start by reusing the upstream BFF; reuse will involve directly using upstream's released images. Code decoration can be used for adding custom logic if needed.
```
bobbravo2, line 33:
```suggestion
- Downstream will start by reusing the upstream BFF; reuse will involve directly using upstream's released images pinned by SHA. 
```

PR #164, opened 2026-09-10
define module NetworkPolicy platform contract

davidebianchi:
## Description
- Add ADR defining NetworkPolicy ownership and lifecycle boundaries for module operators and operands.
- Specify policy requirements, RBAC, validation, reconciliation, cleanup, and platform support limits.
- Document security implications and component-team responsibilities.

Jira task: https://issues.redhat.com/browse/RHOAIENG-90747

## Summary by CodeRabbit

- **Documentation**
  - Defined the NetworkPolicy contract for platform modules, including bundle and operand policy responsibilities.
  - Clarified required selectors, ingress sources, ports, supported peers, and webhook paths.
  - Documented optional egress isolation and requirements for managed egress rules.
  - Established ownership boundaries, including controller responsibilities and operator exclusions.
  - Added shutdown and cleanup requirements for module policies and resources.
  - Expanded validation guidance for policy shape, workload coverage, ownership, recovery, and additive environmental policies.

### DIFF
--- a/architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
+++ b/architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
@@ -1,0 +1,332 @@
+# RHOAI Module NetworkPolicy Platform Contract
+
+|                |            |
+| -------------- | ---------- |
+| Date           | 2026-09-09 |
+| Scope          | Open Data Hub Operator and integrated RHOAI modules |
+| Status         | Draft |
+| Authors        | [Davide Bianchi](@davidebianchi) |
+| Supersedes     | N/A |
+| Superseded by: | N/A |
+| Tickets        | [RHOAIENG-90747](https://issues.redhat.com/browse/RHOAIENG-90747) |
+| Other docs:    | N/A |
+
+## What
+
+This ADR defines the NetworkPolicy platform contract between `opendatahub-operator`, RHOAI module operators, and components. It separates policies delivered with a module operator bundle from policies created for module-managed operands, and assigns one lifecycle manager to each class.
+
+The contract defines required policy manifests, selector ownership, reconciliation behavior, RBAC, validation, and the boundary of platform support. The terms MUST, MUST NOT, SHOULD, and MAY are normative.
+
+## Why
+
+RHOAI modules need predictable, least-privilege network isolation. Module operators and their operands have different deployment lifecycles, so a single NetworkPolicy management model creates ambiguity and can cause competing reconcilers or premature cleanup.
+
+The platform must guarantee that module-bundle policies are deployed and cleaned up with the operator bundle. Each module operator must remain responsible for policies covering its own operands, where it has the required lifecycle and endpoint context.
+
+## Goals
+
+- Define a clear ownership and lifecycle boundary for module-bundle and operand policies.
+- Require least-privilege ingress for module operators and exposed operands.
+- Ensure operand policies are declaratively reconciled and recreated after deletion.
+- Make NetworkPolicy RBAC part of every module's deployment contract.
+- Define bundle and runtime validation responsibilities.
+
+## Non-Goals
+
+- Define one identical ingress rule set for every module.
+- Require `opendatahub-operator` to create or reconcile operand policies.
+- Permit allow-all policies as a substitute for an explicit traffic design.
+- Define controls for secondary networks, host networking, or application-layer authorization.
+- Define or modify labels on workloads not owned by the component.
+
+## How
+
+### Policy classes and ownership
+
+The contract defines two policy classes:
+
+| Policy class | Target | Accountable team | Declared by | Lifecycle manager |
+| --- | --- | --- | --- | --- |
+| Module-bundle policy | Module operator pods, including metrics or webhook endpoints | Component team | Module Helm chart or Kustomize bundle | `opendatahub-operator` through normal module deployment and cleanup |
+| Operand policy | Module-managed operand pods and services | Component team | Module or component controller reconciliation | Responsible module or component controller |
+
+Accountability remains with the component team in both cases. Lifecycle management describes which controller applies and removes the Kubernetes object; it does not transfer policy design responsibility.
+
+### Module-bundle policies
+
+Every module Helm chart or Kustomize bundle MUST contain at least one `networking.k8s.io/v1` `NetworkPolicy` for its module operator. The policy MUST:
+
+- Be rendered by the same Helm chart or Kustomize source as the module operator.
+- Render into the module operator namespace.
+- Select module operator pods using specific, stable labels.
+- Declare explicit `policyTypes`.
+- Declare explicit ingress sources and ports for every allowed path.
+- Avoid empty selectors and allow-all ingress.
+- Work in every supported overlay.
+
+If the module operator accepts no ingress traffic, its policy MUST declare `policyTypes: [Ingress]` with no ingress rules. Multiple policies are allowed, for example separate policies for metrics and webhook traffic.
+
+`opendatahub-operator` installs bundle policies with the module operator resources, manages their rendered state, and removes them during module-resource cleanup. The module operator MUST NOT reconcile the same bundle policy. It may read or watch the object when required, but must not compete with the platform field manager or assign a second controller owner.
+
+### Operand policies
+
+Every module-managed workload that accepts traffic through a Service, webhook, sidecar, user-facing endpoint, or documented pod-to-pod listener MUST be selected by an operand NetworkPolicy controlling that ingress. A module MUST NOT rely only on a default namespace policy.
+
+Expected source restrictions include:
+
+| Endpoint | Required allowed source | Required ports |
+| --- | --- | --- |
+| Webhook | Kubernetes API server | Webhook port only |
+| Metrics | Monitoring collectors | Explicit metrics port(s) only |
+| Gateway-routed API or user endpoint | Approved gateway pods | Advertised endpoint port(s) only |
+| Internal API | Explicit component callers | Declared endpoint port(s) only |
+
+Component teams MUST define and document the concrete target selectors, allowed source peers, and ports for each exposed endpoint. The target `spec.podSelector` MUST use stable labels applied to workloads owned by the component. Source selectors for pod-based callers MUST use stable labels owned by the calling service. Placeholders and unresolved source identities are non-compliant.
+
+Webhook policies MUST permit Kubernetes API-server calls only on the webhook port. Controllers MUST NOT derive selectors from currently running pods or add labels to workloads they do not own.
+
+Static operand policies placed only in a platform Kustomize overlay or Helm chart do not satisfy this contract. Modules using [ODH Platform Utilities](https://github.com/opendatahub-io/odh-platform-utilities) SHOULD use its deploy, dynamic ownership, and garbage-collection actions to implement the reconciliation and lifecycle requirements below.
+
+The following controller-runtime snippets are illustrative. They explain the required ownership and watch semantics but do not prescribe controller setup when the shared framework provides equivalent behavior. For example, a controller responsible for the operand lifecycle could register an ownership watch with:
+
+```go
+return ctrl.NewControllerManagedBy(mgr).
+    For(&Operand{}).
+    Owns(&networkingv1.NetworkPolicy{}).
+    Complete(r)
+```
+
+`Owns` configures the watch but does not establish the Kubernetes owner reference. When the operand can be a valid owner of the namespaced policy, equivalent low-level controller-runtime code sets that reference before creating or applying the policy and propagates any error:
+
+```go
+if err := controllerutil.SetControllerReference(operand, policy, r.Scheme); err != nil {
+    return ctrl.Result{}, err
+}
+```
+
+If a valid owner reference cannot be used, such as when owner and policy scopes differ, the controller MUST use an explicit watch mapping and MUST delete the policy during operand cleanup. In either case, policy deletion MUST enqueue reconciliation.
+
+During normal operand reconciliation, the responsible module or component controller MUST:
+
+- Create policies.
+- Recreate policies after accidental deletion.
+- Update policies when desired configuration changes.
+- Delete policies when the owning operand is removed or disabled.
+- Use declarative apply semantics so manual changes do not permanently diverge from desired state.
+
+The operand policy lifecycle MUST follow the operand lifecycle. The platform MUST NOT create, update, or garbage-collect operand policies.
+
+> **Note:** This contract covers traffic on the primary Kubernetes pod network.
+> Components using secondary network attachments, such as SR-IOV,
+> RDMA/RoCE, or Ray Clusters configured on a secondary interface, MUST document and test
+> applicable network-specific controls. Standard `NetworkPolicy` coverage MUST
+> NOT be assumed for that traffic.
+
+### User-facing operands
+
+User-facing operands are included in the operand-policy contract. The fact that a user creates the operand does not make the user responsible for its NetworkPolicy.
+
+The component controller responsible for integrating the operand's generated workloads and platform networking MUST create and reconcile their NetworkPolicies. This may be the workload controller itself, a dedicated component controller, or a companion platform-integration controller. That controller owns the policy's desired state even when another controller creates the generated Pods, Deployments, StatefulSets, or Services.
+
+The responsible policy controller MUST have NetworkPolicy RBAC, declare ownership or an equivalent watch relationship, and make policy cleanup follow the user-facing operand lifecycle. If an upstream workload controller cannot reconcile the required policy, the RHOAI component integration controller MUST provide that reconciliation. The platform operator MUST NOT become a second reconciler.
+
+User-facing policies MUST cover endpoint paths exposed by the operand, including approved user or gateway ingress, component-to-operand traffic, monitoring, and required platform-service or DNS traffic. A shared policy is acceptable only when its selector and lifecycle remain specific to the supported operand set; a namespace-wide allow policy is not sufficient.
+
+### Policy content requirements
+
+Every operand policy MUST target a specific pod set:
+
+- `spec.podSelector` MUST contain meaningful labels.
+- `podSelector: {}` is prohibited.
+- Stable `app.kubernetes.io/*` labels or an equivalent module-owned label contract SHOULD be used.
+
+Ingress rules MUST be explicit. Each ingress rule MUST contain a non-empty `from` list with specific NetworkPolicy peers and a non-empty `ports` list. Pod-based callers SHOULD be restricted by both namespace and pod selectors. A namespace-only source is acceptable only when that namespace is dedicated to the approved caller and the broader access is documented. These patterns are prohibited:
+
+```yaml
+ingress:
+- {}
+```
+
+```yaml
+from: []
+```
+
+Omitting `from` or `ports`, using an empty list, or otherwise allowing unrestricted ingress is non-compliant.
+
+Workloads handling credentials, tokens, object-storage access, or other sensitive data MUST restrict egress to known destinations. An egress policy MUST select those workloads with a meaningful, stable `podSelector`, include `Egress` in `policyTypes`, and define non-empty `to` and `ports` lists. A dedicated egress policy uses `policyTypes: [Egress]`.
+
+The allowlist MUST contain only destinations required by the workload.
+
+The complete installed policy set MUST permit no other egress for those workloads. In-cluster destinations SHOULD use stable namespace and pod selectors. Destinations represented by IP ranges MUST use the narrowest stable CIDRs available. Because Kubernetes NetworkPolicy cannot select destinations by DNS name, modules using hostname-only or dynamically addressed services MUST document that limitation and any supplemental control.
+
+### Controller RBAC and configuration
+
+The module or component controller ClusterRole or namespace-scoped Role MUST include the NetworkPolicy permissions required for its policy scope:
+
+```yaml
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - networkpolicies
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+```
+
+Permissions MUST be present in the controller's Helm chart or Kustomize manifests. Platform RBAC does not replace controller RBAC. Namespace-scoped permissions SHOULD be used when policies are confined to one namespace. Cluster-scoped permissions are appropriate only when a component intentionally manages policies across namespaces.
+
+Modules MAY expose operand policy configuration through their module or operand CRD. A recommended shape is:
+
+```yaml
+spec:
+  networkPolicy:
+    ingress:
+      enabled: true
+```
+
+For modules exposing services, policy creation MUST be enabled by default. If supported configuration disables policy creation, the deployment no longer complies with this contract. The module MUST document the reduced security posture, retain NetworkPolicy RBAC, and report the disabled protection in component status or another administrator-visible signal.
+
+### Lifecycle semantics
+
+Module-bundle policies follow the module operator resource lifecycle:
+
+1. The platform renders and applies the policy with module operator resources.
+2. Platform field management and Server-Side Apply manage rendered policy changes.
+3. When a module is disabled, the module CR is removed first while the module operator remains available to clean up operands.
+4. After the module CR is gone, the platform removes rendered module resources, including bundle policies.
+
+Operand policies follow the module CR and operand lifecycle:
+
+1. The responsible module or component controller creates them and establishes an owner reference or explicit lifecycle mapping.
+2. The responsible controller recreates them after deletion.
+3. Removing or disabling an operand removes its policies through controller cleanup or verified garbage collection.
+4. The platform does not delete them by rendering the module bundle.
+
+The two lifecycle paths MUST NOT target the same NetworkPolicy object.
+
+### Validation contract
+
+Validation has separate bundle and runtime gates.
+
+#### Bundle validation in `opendatahub-operator`
+
+The platform module-manifest compliance test MUST render every registered module Helm/Kustomize source and verify:
+
+- Every rendered module operator workload is selected by at least one NetworkPolicy.
+- Each policy uses `networking.k8s.io/v1` and kind `NetworkPolicy`.
+- Policy namespace matches the rendered module operator namespace.
+- Each `podSelector` matches its intended rendered workload and no unrelated workload.
+- `policyTypes` is explicit and includes `Ingress` for endpoint-protecting policies.
+- Every allowed ingress rule has explicit sources and ports; a policy with no ingress rules is valid when the selected operator requires no ingress.
+- No empty selector or allow-all ingress pattern is present.
+
+This test validates module-bundle policies only. It MUST NOT claim to validate operand coverage or module-controller ownership.
+
+#### Runtime validation by component teams
+
+Component teams own runtime validation for their traffic paths and policy lifecycle. Module or component integration tests MUST verify operand policies:
+
+- Every documented endpoint has corresponding policy coverage.
+- Target selectors match the intended workloads and no unrelated pods.
+- Source peers and ports match the documented callers and endpoints.
+- Deleting a policy causes the responsible controller to recreate it.
+- Removing or disabling an operand leaves no policy created for that operand after reconciliation and cleanup complete.
+- Manual policy changes are reconciled to desired state.
+
+Where NetworkPolicy enforcement is available, component integration tests MUST also verify that unauthorized cross-namespace ingress is denied, approved module, gateway, monitoring, and webhook traffic continues to work, and sensitive-workload egress allows documented destinations while denying an unapproved destination.
+
+Traffic-behavior tests MUST evaluate the complete installed policy set because multiple NetworkPolicies are additive. The preferred test uses a deny-all baseline and verifies only explicitly supported traffic paths.
+
+### Platform support boundary
+
+The platform supports this contract by:
+
+- Rendering and applying module-bundle policies with module operator resources.
+- Validating bundle-policy presence and shape during module compliance checks.
+- Preserving NetworkPolicy resources through Helm/Kustomize rendering and deployment.
+- Passing platform namespace configuration required by module policies.
+- Cleaning up bundle policies with other rendered module resources.
+
+It does not:
+
+- Create operand policies on behalf of modules.
+- Add platform ownership to operand policies.
+- Use module garbage collection to delete operand policies.
+- Treat one bundle policy as sufficient operand-policy compliance.
+
+## Alternatives
+
+### Platform manages all NetworkPolicies
+
+Have `opendatahub-operator` create and reconcile both module-bundle and operand policies.
+
+This centralizes validation and lifecycle handling, but the platform operator lacks complete knowledge of module-created operands, generated endpoints, and module-specific configuration. It also expands platform RBAC and creates coupling between platform and component release cycles.
+
+### Module operators manage all NetworkPolicies
+
+Have each module operator also manage the policy for its own operator pods.
+
+This gives one controller all policy context for a module, but conflicts with the platform's bundle deployment and cleanup lifecycle. It can produce competing field managers and leave operator pods unprotected when the module operator is unavailable.
+
+### Static operand policies in the module bundle
+
+Ship operand policies with the module Helm or Kustomize source and let `opendatahub-operator` apply them.
+
+This is simple for fixed workloads but cannot reliably follow user-created operands, generated namespaces, endpoint configuration, or operand deletion. It also gives the platform deployer lifecycle responsibility for resources whose desired state is known only to the component controller.
+
+### Namespace-wide allow policies
+
+Use broad namespace-level policies and rely on service authentication for finer control.
+
+This reduces component effort but violates least privilege, permits unintended cross-component traffic, and is insufficient as a network isolation contract.
+
+## Security and Privacy Considerations
+
+- NetworkPolicy is defense in depth; it does not replace application authentication, authorization, or encryption.
+- Explicit selectors, sources, and ports reduce unintended exposure and cross-namespace access.
+- Policy changes can deny required control-plane, monitoring, webhook, or DNS traffic; component tests MUST cover approved traffic paths.
+- Sensitive workloads MUST restrict egress with explicit `Egress` rules covering only required DNS, Kubernetes API, and documented service destinations.
+- NetworkPolicies are additive. A broad policy selecting the same pods can expand allowed traffic, so validation MUST evaluate the complete installed policy set.
+- Policy configuration that disables protection MUST be documented as a reduced security posture.
+- This contract does not introduce collection or persistence of application data.
+
+## Risks
+
+- Incorrect selectors or platform namespace assumptions can break module availability.
+- A cluster can use a CNI that does not support NetworkPolicy. Kubernetes may accept policy objects without enforcing isolation. Because the platform supports both OCP and generic Kubernetes, CNI capabilities cannot be assumed uniformly.
+- Hostname-only or dynamically addressed egress destinations cannot be represented precisely by standard NetworkPolicy.
+- An unrelated broad NetworkPolicy can weaken isolation because allowed traffic is the union of all policies selecting a pod.
+- A module that omits operand reconciliation can appear compliant at bundle-render time while leaving runtime endpoints exposed.
+
+## Stakeholder Impacts
+
+| Group | Key Contacts | Date | Impacted? |
+| --- | --- | --- | --- |
+| ODH Platform / Operator | | 2026-09-09 | Yes |
+| Module and component teams | | 2026-09-09 | Yes |
+| Model Serving | | 2026-09-09 | Yes |
+| Data Science Pipelines | | 2026-09-09 | Yes |
+| Distributed Workloads | | 2026-09-09 | Yes |
+| Model Registry | | 2026-09-09 | Yes |
+| Dashboard and workbench teams | | 2026-09-09 | Yes |
+| Security and release engineering | | 2026-09-09 | Yes |
+
+Component teams must add bundle policies, operand reconciliation, RBAC, and runtime tests. The platform team must run bundle compliance tests, and preserve the lifecycle boundary.
+
+## References
+
+- [RHOAIENG-90747](https://issues.redhat.com/browse/RHOAIENG-90747)
+- [Kubernetes NetworkPolicies](https://kubernetes.io/docs/concepts/services-networking/network-policies/)
+- [Kubernetes NetworkPolicy API](https://kubernetes.io/docs/reference/kubernetes-api/networking/network-policy-v1/)
+- [ODH Platform Utilities](https://github.com/opendatahub-io/odh-platform-utilities)
+
+## Reviews
+
+| Reviewed by | Date | Notes |
+| --- | --- | --- |
+| | | |


architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,303 @@
...
+The platform module-manifest compliance test MUST render every registered module Helm/Kustomize source and verify:
+
+- At least one NetworkPolicy exists per module bundle.
+- Each policy uses `networking.k8s.io/v1` and kind `NetworkPolicy`.
+- Policy namespace matches the rendered module operator namespace.
+- `podSelector` selects the rendered module operator Deployment.
+- `policyTypes` is explicit and includes `Ingress` for endpoint-protecting policies.
+- Every ingress rule has explicit sources and ports.
+- No empty selector or allow-all ingress pattern is present.
+
```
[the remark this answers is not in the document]
davidebianchi: Not a goal for this ADR

matzew (commented):
Feedback from the **mcp-lifecycle-operator (OCPMCPLO)** team, reviewed against our
per-MCPServer operand NetworkPolicy reconciler on `upstream/main`.

Overall **+1** on the direction and the explicit ownership/lifecycle split - it
matches how the operator already reconciles operand policies today
(owner-referenced, GC-following, declaratively applied).

We read the ADR as making our current **default** (when a user sets no network
restrictions) non-compliant in two ways - allow-all ingress source and allow-all
egress - which we will fix under **RHOAIENG-93738**. Before we can implement a
*compliant* default, we need clarification on two points specific to
gateway-routed, proxy-style operands (inline questions on the relevant lines):

1. How to express the "approved gateway pods" default source when the operator
   does not know the gateway at NP-creation time (line 81).
2. What a compliant default egress looks like for operands that reach
   hostname-only / dynamically addressed external endpoints (lines 156 / 160).

Plus one smaller question on whether an audit event / log satisfies the
"administrator-visible signal" requirement vs. a status condition (line 192).

Thanks for writing this up - the contract is clear and useful.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+Every module-managed workload that accepts traffic through a Service, webhook, sidecar, user-facing endpoint, or documented pod-to-pod listener MUST be selected by an operand NetworkPolicy controlling that ingress. A module MUST NOT rely only on a default namespace policy.
+
+Expected source restrictions include:
+
+| Endpoint | Required allowed source | Required ports |
+| --- | --- | --- |
+| Webhook | Kubernetes API server | Webhook port only |
+| Metrics | Monitoring collectors | Explicit metrics port(s) only |
+| Gateway-routed API or user endpoint | Approved gateway pods | Advertised endpoint port(s) only |
```
matzew, line 81: For the mcp-lifecycle-operator, MCPServer operands map to this row
(gateway-routed / user endpoint). The operand NetworkPolicy is reconciled
per-MCPServer by the operator, which does not reliably know the approved
gateway's identity at NP-creation time.

Could the ADR specify the sanctioned way to express "approved gateway pods" as a
**default** source - e.g. a well-known gateway namespace/pod label contract the
platform guarantees? Without that, a module operator cannot produce a compliant
default source and is forced back to allow-all. This is the main blocker for us
to flip our default to least-privilege.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+
+### Policy content requirements
+
+Every operand policy MUST target a specific pod set:
+
+- `spec.podSelector` MUST contain meaningful labels.
+- `podSelector: {}` is prohibited.
+- Stable `app.kubernetes.io/*` labels or an equivalent module-owned label contract SHOULD be used.
+
+Ingress rules MUST be explicit. Each ingress rule MUST contain a non-empty `from` list with specific NetworkPolicy peers and a non-empty `ports` list. Pod-based callers SHOULD be restricted by both namespace and pod selectors. A namespace-only source is acceptable only when that namespace is dedicated to the approved caller and the broader access is documented. These patterns are prohibited:
```
matzew, line 143: Concrete impact for us: when a user sets no `Spec.Network`, the
mcp-lifecycle-operator operand policy today emits an ingress rule with `ports`
set but an empty `from` (allow-all source) - i.e. the prohibited pattern above.

We read this ADR as requiring us to change that default (tracked under
RHOAIENG-93738). Confirming that reading is correct, and that there is no
"unset = allow-all is acceptable if audited" exception.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+- {}
+```
+
+```yaml
+from: []
+```
+
+Omitting `from` or `ports`, using an empty list, or otherwise allowing unrestricted ingress is non-compliant.
+
+Workloads handling credentials, tokens, object-storage access, or other sensitive data MUST restrict egress to known destinations. An egress policy MUST select those workloads with a meaningful, stable `podSelector`, include `Egress` in `policyTypes`, and define non-empty `to` and `ports` lists. A dedicated egress policy uses `policyTypes: [Egress]`.
```
matzew, line 156: MCP servers proxy user tokens/credentials to upstream MCP endpoints, so they
fall under this sensitive-workload egress requirement. Our current default
(no `Spec.Network`) emits a single empty egress rule = allow-all egress.

We plan to make this least-privilege under RHOAIENG-93738 - see the DNS-name
limitation question below, which affects what a compliant *default* can even look
like for a proxy-style operand.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+
+The allowlist MUST contain only destinations required by the workload.
+
+The complete installed policy set MUST permit no other egress for those workloads. In-cluster destinations SHOULD use stable namespace and pod selectors. Destinations represented by IP ranges MUST use the narrowest stable CIDRs available. Because Kubernetes NetworkPolicy cannot select destinations by DNS name, modules using hostname-only or dynamically addressed services MUST document that limitation and any supplemental control.
```
matzew, line 160: This carve-out is critical for proxy-style operands. MCP servers commonly egress
to hostname-only / dynamically addressed external MCP endpoints that are unknown
at NP-creation time and cannot be expressed as CIDRs or selectors.

Can the ADR clarify what a *compliant default* is for this class - e.g. is
"document the limitation + restrict to DNS + user-declared `EgressTo`" sufficient,
or is a specific supplemental control expected? This determines whether our
default can be least-privilege without breaking connectivity.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+Modules MAY expose operand policy configuration through their module or operand CRD. A recommended shape is:
+
+```yaml
+spec:
+  networkPolicy:
+    ingress:
+      enabled: true
+```
+
+For modules exposing services, policy creation MUST be enabled by default. If supported configuration disables policy creation, the deployment no longer complies with this contract. The module MUST document the reduced security posture, retain NetworkPolicy RBAC, and report the disabled protection in component status or another administrator-visible signal.
```
matzew, line 192: When ingress/egress restrictions are absent, the mcp-lifecycle-operator currently
signals the reduced posture via a structured audit event + operator log rather
than a condition on the MCPServer status.

Does an audit event / log qualify as an "administrator-visible signal" here, or
does the ADR intend a status condition specifically? Happy to add a status
condition if that is the expectation.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+| Tickets        | [RHOAIENG-90747](https://issues.redhat.com/browse/RHOAIENG-90747) |
+| Other docs:    | N/A |
+
+## What
+
+This ADR defines the NetworkPolicy platform contract between `opendatahub-operator`, RHOAI module operators, and components. It separates policies delivered with a module operator bundle from policies created for module-managed operands, and assigns one lifecycle manager to each class.
+
+The contract defines required policy manifests, selector ownership, reconciliation behavior, RBAC, validation, and the boundary of platform support. The terms MUST, MUST NOT, SHOULD, and MAY are normative.
+
+## Why
```
sefroberg, line 20: Do we also want to tie into Red Hat's overall strategy for tighting up network communication  security here with the release of OCP 5?

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+
+## Risks
+
+- Incorrect selectors or platform namespace assumptions can break module availability.
+- A cluster can use a CNI that does not support NetworkPolicy. Kubernetes may accept policy objects without enforcing isolation. Because the platform supports both OCP and generic Kubernetes, CNI capabilities cannot be assumed uniformly.
+- Hostname-only or dynamically addressed egress destinations cannot be represented precisely by standard NetworkPolicy.
+- An unrelated broad NetworkPolicy can weaken isolation because allowed traffic is the union of all policies selecting a pod.
+- A module that omits operand reconciliation can appear compliant at bundle-render time while leaving runtime endpoints exposed.
+
+## Stakeholder Impacts
```
sefroberg, line 306: This does not seem like a complete list. Is this intended to map to the component teams?

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+- Require least-privilege ingress for module operators and exposed operands.
+- Ensure operand policies are declaratively reconciled and recreated after deletion.
+- Make NetworkPolicy RBAC part of every module's deployment contract.
+- Define bundle and runtime validation responsibilities.
+
+## Non-Goals
+
+- Define one identical ingress rule set for every module.
+- Require `opendatahub-operator` to create or reconcile operand policies.
+- Permit allow-all policies as a substitute for an explicit traffic design.
```
lphiri, line 38: This is going to be tricky for shared services such as MLFlow,  Feast, Model registry that do not know at NP creation time where connections are going to come from.
Do we need to have some guidance/requirements for adding support for service admin defined rules?
Dynamic rules in general are going to be problematic without "allow-all" policies.

We must also consider module to module communication rules/expectations explicitly.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
lphiri: +1  to allowing for user-declared Egress for this one.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
lphiri: MLFlow supports use configured NP rules (networkPolicyAdditionalEgressRules)  . I wonder if we can adopt this as a general pattern:
 https://github.com/opendatahub-io/mlflow-operator

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+### Platform support boundary
+
+The platform supports this contract by:
+
+- Rendering and applying module-bundle policies with module operator resources.
+- Validating bundle-policy presence and shape during module compliance checks.
+- Preserving NetworkPolicy resources through Helm/Kustomize rendering and deployment.
+- Passing platform namespace configuration required by module policies.
+- Cleaning up bundle policies with other rendered module resources.
+
```
lphiri, line 254: Should we consider multi-tenancy hooks as part this contract?
For example could the platform inject/or advertize which namespaces are part of the same tenant in the cluster?

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
ugiordan, line 38: Could we make the boundary of this ADR a bit more explicit?

My understanding is that this contract defines who owns each NetworkPolicy and how policies should be reconciled and tested on the primary pod network. It does not provide the complete multi-tenancy security model.

In particular, it does not by itself guarantee:

- non-overridable cross-tenant isolation, because standard NetworkPolicies are additive;
- protection of secondary interfaces such as Multus, SR-IOV, or RDMA;
- portable enforcement for hostNetwork workloads;
- authentication, authorization, application-level tenant data isolation, or encryption;
- isolation of shared-service data after network access is granted.

It would help to state this directly and, where possible, point to the follow-up work that owns those areas. That would prevent readers from treating compliance with this ADR as proof of complete tenant isolation.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+This is simple for fixed workloads but cannot reliably follow user-created operands, generated namespaces, endpoint configuration, or operand deletion. It also gives the platform deployer lifecycle responsibility for resources whose desired state is known only to the component controller.
+
+### Namespace-wide allow policies
+
+Use broad namespace-level policies and rely on service authentication for finer control.
+
+This reduces component effort but violates least privilege, permits unintended cross-component traffic, and is insufficient as a network isolation contract.
+
+## Security and Privacy Considerations
```
ugiordan, line 288: Please state explicitly that compliance with this contract proves component-level NetworkPolicy ownership and tested L3/L4 traffic restrictions on supported networks. It does not prove complete tenant isolation, application authorization, data partitioning, or encryption.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
ugiordan: I think the important constraint is that these rules should extend a secure controller-owned baseline rather than replace it. The API should keep ingress and egress separate, restrict who can configure additional rules, reject empty peers or omitted ports, and report the resulting policy in the status.

For module-to-module traffic, I think each side should remain responsible for its own direction: the caller manages egress and the destination manages ingress. That avoids one module changing policies owned by another module.

We will still need an integration test for the final policy set because each rule might look correct on its own while the combined result is broader than expected.

davidebianchi: The multi-tenancy security model will be provided by the multi tenancy ADR. This is to define how the NetworkPolicies should be defined by operators in the platform.
There are discussion on how to update this contract based on how it could work together with multi-tenancy, because actually it needs some refinement (e.g. probably we should make the egress policies rules less strict).

> That would prevent readers from treating compliance with this ADR as proof of complete tenant isolation.

This ADR is not about tenancy, but the single component policies management and standard guidelines. It never mentions multi-tenancy, and should be integrated in components before multi tenancy work.

ugiordan: Yes, that makes sense. I agree that the tenant security model belongs in the multi-tenancy ADR.

My point is that this contract still affects multi-tenancy because tenant-level isolation will depend on how component policies are defined and combined. A short sentence clarifying that relationship would be enough for me.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
ugiordan: Could we address this in the shared library with an API similar to [nais/fqdn-policy](https://github.com/nais/fqdn-policy)? A component could declare the external FQDNs and ports it needs, and the library could reconcile the corresponding policy while keeping egress deny-by-default.

On OpenShift, the implementation could use the OVN-Kubernetes EgressFirewall. It supports DNS names and would allow MCP endpoints without opening all egress.

[NPEP-133](https://network-policy-api.sigs.k8s.io/npeps/npep-133/) is the relevant upstream proposal for FQDN-based egress.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
ugiordan: I assume this list represents the components already onboarded as modules. I agree that it should be extended to include all components covered by this contract, or explicitly labeled as an initial list.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
ugiordan: I think this is a useful boundary. The multi-tenancy ADR should define how namespaces are assigned to tenants and how that assignment is protected. This ADR can define how component policies consume that information.

For example, the platform could provide a protected tenant label or an equivalent API. Module controllers would use it in their selectors, but they should not infer tenant membership from arbitrary labels or change the tenant identity themselves. The contract should also define what happens when the identity is missing or changes. It should never fall back to a broad allow.

The platform-provided identity would make policy generation consistent, but it would not by itself enforce a hard tenant boundary. Standard NetworkPolicies are additive, so another policy selecting the same pods could add broader access. Hard isolation therefore also requires protected policy permissions, admission control, or a higher-precedence policy mechanism.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+    return ctrl.Result{}, err
+}
+```
+
+If a valid owner reference cannot be used, such as when owner and policy scopes differ, the controller MUST use an explicit watch mapping and MUST delete the policy during operand cleanup. In either case, policy deletion MUST enqueue reconciliation.
+
+During normal operand reconciliation, the responsible module or component controller MUST:
+
+- Create policies.
+- Recreate policies after accidental deletion.
```
ugiordan, line 117: One gap I still see is what happens when a required policy cannot be restored. The ADR says the controller must recreate it after deletion, but it does not define how failures should be handled.

Could we define retry and error behavior for forbidden, conflict, and other API errors, and require a durable degraded condition or equivalent signal while the policy is missing? The tests should cover both the failure and the recovery path. `Owns` only triggers reconciliation; it does not prove that restoration succeeded.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+- Every documented endpoint has corresponding policy coverage.
+- Target selectors match the intended workloads and no unrelated pods.
+- Source peers and ports match the documented callers and endpoints.
+- Deleting a policy causes the responsible controller to recreate it.
+- Removing or disabling an operand leaves no policy created for that operand after reconciliation and cleanup complete.
+- Manual policy changes are reconciled to desired state.
+
+Where NetworkPolicy enforcement is available, component integration tests MUST also verify that unauthorized cross-namespace ingress is denied, approved module, gateway, monitoring, and webhook traffic continues to work, and sensitive-workload egress allows documented destinations while denying an unapproved destination.
+
+Traffic-behavior tests MUST evaluate the complete installed policy set because multiple NetworkPolicies are additive. The preferred test uses a deny-all baseline and verifies only explicitly supported traffic paths.
```
ugiordan: What does “complete installed policy set” mean here? Does the qualification baseline include component policies, platform policies, and any supported higher-precedence policies?

This matters because NetworkPolicies are additive. Another policy selecting the same pods can broaden access even when the component’s own policy is least-privilege.

Could the ADR require the test evidence to list every applicable policy and evaluate their combined effect? Policies created outside the qualification baseline should be reported as environmental overrides rather than attributed to the component.

It would also help to clarify whether the deny-all baseline is mandatory or simply one supported test strategy.

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,332 @@
...
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+```
+
+Permissions MUST be present in the controller's Helm chart or Kustomize manifests. Platform RBAC does not replace controller RBAC. Namespace-scoped permissions SHOULD be used when policies are confined to one namespace. Cluster-scoped permissions are appropriate only when a component intentionally manages policies across namespaces.
```
ugiordan:
```suggestion
Permissions MUST be present in the controller's rendered Helm or Kustomize manifests. Platform RBAC does not replace controller RBAC. Namespace-scoped permissions SHOULD be used when policies are confined to one namespace. Cluster-scoped permissions are appropriate only when a component intentionally manages policies across namespaces. Such a component MUST define how namespaces enter its authority set, for example through an owned resource or a platform-owned label protected by RBAC or admission, and MUST reject reconciliation for namespaces outside that set. Bundle validation MUST verify the rendered Role or ClusterRole, its RoleBinding or ClusterRoleBinding, the controller ServiceAccount, and the intended namespace scope. Qualification testing MUST verify that the installed controller can perform the required NetworkPolicy operations using its actual ServiceAccount and covers both an authorized and an unauthorized namespace.

**Review by ugiordan — CHANGES_REQUESTED** [review]

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 6 above)
Comment (ugiordan) [inline architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: What exactly happens when policy protection is disabled? The ADR does not define who is allowed to disable it, whether an existing policy is removed or simply no longer reconciled, how re-enabling works, or how the reduced posture is reported.

Should we explicitly call that, like:
```suggestion
For modules exposing services, policy creation MUST be enabled by default. Disabling it MUST be restricted to an explicitly authorized administrator and MUST have deterministic disable and re-enable behavior. The module MUST report a durable reduced-posture condition while protection is disabled, retain NetworkPolicy RBAC, and restore the policy before reporting protection as enabled again. Disabling a component policy MUST NOT remove or weaken any higher-precedence platform policy.

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md [hunk 15]
```diff
@@ -0,0 +1,332 @@
...
+- Every allowed ingress rule has explicit sources and ports; a policy with no ingress rules is valid when the selected operator requires no ingress.
+- No empty selector or allow-all ingress pattern is present.
+
+This test validates module-bundle policies only. It MUST NOT claim to validate operand coverage or module-controller ownership.
+
+#### Runtime validation by component teams
+
+Component teams own runtime validation for their traffic paths and policy lifecycle. Module or component integration tests MUST verify operand policies:
+
+- Every documented endpoint has corresponding policy coverage.
```
ugiordan, line 234: We should define the evidence that the components need to provide to demonstrate this. We could require an endpoint inventory mapping each exposed workload to the actual selectors, ports, routes, and owning policy, instead of relying only on static manifests.

```suggestion
- Every documented endpoint has corresponding policy coverage, proven by an endpoint inventory that maps each workload and Service to its final pod labels, Service selector, protocol, destination port, route or gateway exposure, sidecars, and owning NetworkPolicy. Generated and user-created workloads MUST be checked against the runtime inventory, not only static manifests.

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md [hunk 16]
```diff
@@ -0,0 +1,332 @@
...
+The responsible policy controller MUST have NetworkPolicy RBAC, declare ownership or an equivalent watch relationship, and make policy cleanup follow the user-facing operand lifecycle. If an upstream workload controller cannot reconcile the required policy, the RHOAI component integration controller MUST provide that reconciliation. The platform operator MUST NOT become a second reconciler.
+
+User-facing policies MUST cover endpoint paths exposed by the operand, including approved user or gateway ingress, component-to-operand traffic, monitoring, and required platform-service or DNS traffic. A shared policy is acceptable only when its selector and lifecycle remain specific to the supported operand set; a namespace-wide allow policy is not sufficient.
+
+### Policy content requirements
+
+Every operand policy MUST target a specific pod set:
+
+- `spec.podSelector` MUST contain meaningful labels.
+- `podSelector: {}` is prohibited.
```
ugiordan: Does the prohibition on `podSelector: {}` also apply to default-deny policies?

For example, this is a valid namespace-wide default-deny policy:

```
spec:
  podSelector: {}
  policyTypes:
  - Ingress
  ingress: []
```

It selects every pod but allows no ingress.

By contrast, this is an allow-all policy:
```
spec:
  podSelector: {}
  policyTypes:
  - Ingress
  ingress:
  - {}
```

It selects every pod and allows traffic from every source on every port.

The contract should scope the prohibition to component allow policies and explicitly allow documented default-deny policies. The validator should still reject wildcard rules such as `ingress: [{}]`.

```suggestion
- An empty `podSelector` MUST NOT be used for a component allow policy. It MAY be used for a documented namespace-wide default-deny policy. Validation MUST distinguish an empty ingress list, which denies ingress, from an empty ingress rule such as `ingress: [{}]`, which allows all sources and ports.

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md [hunk 17]
```diff
@@ -0,0 +1,332 @@
...
+
+Every module Helm chart or Kustomize bundle MUST contain at least one `networking.k8s.io/v1` `NetworkPolicy` for its module operator. The policy MUST:
+
+- Be rendered by the same Helm chart or Kustomize source as the module operator.
+- Render into the module operator namespace.
+- Select module operator pods using specific, stable labels.
+- Declare explicit `policyTypes`.
+- Declare explicit ingress sources and ports for every allowed path.
+- Avoid empty selectors and allow-all ingress.
+- Work in every supported overlay.
```
ugiordan:
```suggestion
- Work in every supported RHOAI/ODH and xKS deployment variant. Each variant MUST be tested, and any platform-specific exceptions MUST be documented.

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 12 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:117]: I think it should be as for the other resources, and if there is a failure it should be raised as the component is not ready in DSC and in the component CR. Would it be enough as addition?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 13 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: Interesting point. It should mean all the network policies in the specific namespaces. For example, actually it means should be tested with all components in DSC managed, since all are installed in the same namespace and could be enough only one allow all policy scoped to all pods in namespace to break the contract.

In the future, when each module will have its own namespace, the complete installed policy should mean the module and all its operands.

For the deny-all, another thing to decide

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 12 above)
Comment (ugiordan) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:117]: Yep!

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 13 above)
Comment (ugiordan) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: That helps. Could we make this scope explicit in the ADR?

For the current deployment model, “complete installed policy set” should mean every policy that can affect the tested pods in the shared DSC namespace. This includes policies owned by the platform, other components, or operands, as well as any supported cluster-scoped policy that applies to those pods. The test should not look only at the policy owned by the component under test.

When modules move to dedicated namespaces, the same rule should apply to the module namespace and its operands. The test evidence should list the applicable policies and evaluate their combined result.


For the deny-all baseline, I think it can remain one supported test strategy rather than a mandatory object. However, every qualification must still include positive and negative traffic tests and must demonstrate that unauthorized ingress and egress are denied.

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 3 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: Yes, the ingress pod or at least namespace should be defined for the network policies.
Do you think would be a problem? If it is something defined from the user, maybe should be exposed a clear field to disable the specific ingress policy, or should we require in case for the user to manually create the network policy?
@lphiri what do you think here?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 2 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: So should be present a standard API to pass the gateway pod information, or a specific label which can be used for the contract?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 5 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: I updated this ADR to treat egress isolation as optional.

Actually, it states that:
When dynamic destinations cannot be represented by standard NetworkPolicy, the component may omit Egress or use a separately governed network-specific control. It must not emit an empty allow-all egress rule. When egress is managed, peers and ports must be explicit.

Optional egress absence does not require a degraded status. Required ingress-policy creation or restoration failures must be reported through the component CR and DSC status.

A follow-up ADR could introduce something like the proposed https://github.com/nais/fqdn-policy, what do you think?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 6 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: I think about a status condition which is more visible, also more in this case since, if I well understood, the network policies should be defined by the user. Is it correct?
Btw, not sure if it should be a requirement or a raccomandation. @lphiri?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 7 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:20]: Added a note in the why section. Is it ok?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 8 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:310]: Since a list would be old in a very short time, I tried with this: https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes#diff-e381837331d816e0fbfd6c163ecad5d8ba7d54eb149fe3bad4c687a8ba5e4f2fR314

WDYT?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 9 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:40]: I added with the last commit something about the module-to-module traffic, can you take a look?
@lphiri now I state that egress are not required, instead ingress are still required. Is it ok, or we should also have some specific case for ingress?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 10 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:257]: I set as non-goal of this ADR: `- Define the multi-tenancy security model.`

Btw, an API to inject some platform-wide inject for network policy could makes sense also in this ADR (not specifically for multi-tenancy, but that can be reused there). Should we do something related here, or it is something for the multi-tenancy ADR?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 11 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:291]: Added here: https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes#diff-e381837331d816e0fbfd6c163ecad5d8ba7d54eb149fe3bad4c687a8ba5e4f2fR297

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 12 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:117]: Added here: https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes#diff-e381837331d816e0fbfd6c163ecad5d8ba7d54eb149fe3bad4c687a8ba5e4f2fR238

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 13 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: The ADR now requires qualification evidence to evaluate every NetworkPolicy selecting the tested pods, record environmental overrides, and test the combined policy effect. A deny-all baseline is not mandatory.

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 6 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: Removed the possibility to disable them, at least for now. In case, it would always be possible for the cluster admin to set allow_all NP in the namespaces

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 15 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:234]: Added this point, is it ok? https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes/88557a5b99432692da05fd37d97d169694d15f3c#diff-e381837331d816e0fbfd6c163ecad5d8ba7d54eb149fe3bad4c687a8ba5e4f2fR234

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 14 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: Added some additional check on the bundle

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 16 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: Explicitly added that default deny policy is allowed.

Btw, it is not safe to be used until all components uses the same namespace. Should be also added this part?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 17 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: Rephrased a little, something like: `Be tested in the module's supported deployment configuration.`

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 9 above)
Comment (lphiri) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:40]: @davidebianchi I think this is fine.  We can revisit if needed.

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 8 above)
Comment (sefroberg) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md:310]: I think that will work. You can mark this resolved.

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 16 above)
Comment (davidebianchi) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: I was rethinking about this part. Also the deny policies should not have empty pod selector, as it could break other workload running in the namespace (e.g. the application namespace, but also when each component will have a new specific namespace).
The deny-all namespace policy could be an enforcement done by the platform operator for example. But since it would cover all the workloads, I'd prefer to remove the possibility here.
@ugiordan what do you think about?

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md (hunk 6 above)
Comment (lphiri) [reply architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md]: @davidebianchi I think a status would be helpful here, but maybe as a SHOULD, rather than MUST

**Review by lphiri — APPROVED** [review]
@davidebianchi  this looks good overall.
I suspect that while we still have shared module namespaces will need to revisit this ADR and tweak some rules.
Lets have @ugiordan  review the requested changes then we can merge.

**Comment by lburgazzoli** [comment]
@davidebianchi we should probably extend/configure https://github.com/stackrox/kube-linter to calidate the NP

File: architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md [hunk 18]
```diff
@@ -0,0 +1,331 @@
...
+
+Every module Helm chart or Kustomize bundle MUST contain at least one `networking.k8s.io/v1` `NetworkPolicy` for its module operator. The policy MUST:
+
+- Be rendered by the same Helm chart or Kustomize source as the module operator.
+- Render into the module operator namespace.
+- Select module operator pods using specific, stable labels.
+- Declare explicit `policyTypes`.
+- Declare explicit ingress sources and ports for every allowed path.
+- Use specific pod selectors for operator allow policies; a documented namespace-wide default-deny policy is permitted.
+- Be tested in the module's supported deployment configuration.
```
ugiordan, line 70: What does “supported deployment configuration” mean here? Since this contract is intended to cover RHOAI, ODH, OCP, and xKS variants, could we require the qualification evidence to name each supported variant and document any platform-specific limitations or exceptions?

Otherwise, a passing test in one overlay could be interpreted as proof that the contract works everywhere.

ugiordan:
```suggestion
- Be tested in every supported product and platform variant. Qualification evidence MUST identify the variant tested and document any platform-specific limitations or approved exceptions.
```

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,331 @@
...
+#### Runtime validation by component teams
+
+Component teams own runtime validation for their traffic paths and policy lifecycle. Module or component integration tests MUST verify operand policies:
+
+- Every documented endpoint has corresponding policy coverage.
+- Test evidence identifies each workload selector, Service and port, exposure path, allowed peer, and owning NetworkPolicy. Generated and user-created workloads are checked at runtime, not only from static manifests.
+- Allow-policy target selectors match the intended workloads and no unrelated pods. A documented namespace-wide default-deny policy MAY select every pod in its namespace.
+- Source peers and ports match the documented callers and endpoints.
+- Deleting a policy causes the responsible controller to recreate it.
+- Failure to create or restore a required policy reports the component as not ready or degraded in both its component CR and the DSC component status; the status clears after recovery.
```
ugiordan, line 239:
```suggestion
- When creation or restoration of a required policy fails, the responsible controller MUST retry recoverable errors and report the component as not ready or degraded in both its component CR and the DSC component status. The condition MUST clear only after the desired policy has been restored successfully.
```

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,331 @@
...
+  - get
+  - list
+  - patch
+  - update
+  - watch
+```
+
+Permissions MUST be present in each controller's rendered Helm or Kustomize manifests. Platform RBAC does not replace controller RBAC. Namespace-scoped permissions SHOULD be used when policies are confined to one namespace. Each controller using cluster-scoped permissions MUST define its NetworkPolicy authority set and reject reconciliation outside it. Runtime validation for each controller MUST use its actual ServiceAccount and cover an authorized namespace and a namespace outside that authority set.
+
+Modules MAY expose policy configuration through a resource appropriate to their use case. This ADR does not prescribe a shared Custom Resource or schema. Configured ingress or egress rules MUST be additive to the controller-owned policy baseline and meet the selector, peer, and port requirements in this ADR. A NetworkPolicy not reconciled by the responsible controller is an environmental override: it neither replaces the required component policy nor becomes part of its lifecycle.
```
ugiordan, line 188: The additive behavior is clear, but the contract does not say who is allowed to configure these additional rules. A user-controlled rule could still broaden the effective policy set even when the controller-owned baseline is restrictive. We can write something like:

```suggestion
Modules MAY expose policy configuration through a resource appropriate to their use case. This ADR does not prescribe a shared Custom Resource or schema. Any configuration that adds ingress or egress rules MUST be writable only by authorized administrators or platform configuration. The rules MUST be additive to the controller-owned policy baseline, MUST meet the selector, peer, and port requirements in this ADR, and MUST NOT remove or weaken required protection. A reduced effective posture MUST be reported through an administrator-visible status or audit signal.
```

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
```diff
@@ -0,0 +1,331 @@
...
+- Define bundle and runtime validation responsibilities.
+
+## Non-Goals
+
+- Define one identical ingress rule set for every module.
+- Require `opendatahub-operator` to create or reconcile operand policies.
+- Permit allow-all policies as a substitute for an explicit traffic design.
+- Require or prescribe egress isolation for every component.
+- Define a shared API or Custom Resource for NetworkPolicy configuration.
+- Define the multi-tenancy security model.
```
ugiordan, line 43:
```suggestion
- Define the multi-tenancy security model. The multi-tenancy ADR owns tenant identity and namespace membership. Component policy controllers MAY consume a protected, platform-provided tenant identity when building tenant-aware selectors, but MUST NOT infer or modify tenant membership.
```

davidebianchi: Updated in https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes/0799faa17565f05422cb7ed751b8657ac6f416a1

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
davidebianchi: Updated in https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes/0799faa17565f05422cb7ed751b8657ac6f416a1

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
davidebianchi: Updated in https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes/0799faa17565f05422cb7ed751b8657ac6f416a1

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
davidebianchi: Updated in https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes/0799faa17565f05422cb7ed751b8657ac6f416a1

architecture-decision-records/operator/ODH-ADR-Operator-0016-networkpolicy-platform-contract.md
davidebianchi: Updated in https://github.com/opendatahub-io/architecture-decision-records/pull/164/changes/0799faa17565f05422cb7ed751b8657ac6f416a1

ugiordan (approved):

PR #169, opened 2026-09-17
ODH-ADR-MS-0006-maas-flow-control: Subscription-Based Flow Control for MaaS

pierDipi:
Introduce requestPriority independently of subscription-selection priority so operators can configure inference scheduling without changing which subscription is selected for a request.

Use the AI tenant as the fairness identity and the selected subscription as the logical objective. This enables configured fairness policies to share dispatch opportunities within a priority band while subscriptions express different scheduling tiers.

Have the subscription controller discover active pools through LLMInferenceService status and reconcile one InferenceObjective per subscription/pool pair. Inject matching objective names only after authentication and authorization so clients cannot assert their own tier.

Leave objective resources absent when requestPriority is unset, relying on the scheduler's default priority of zero. Always inject trusted fairness and objective headers in both canonical and legacy forms. Preserve explicit zero and negative priorities, and reconcile changes without recreating subscriptions.

Detailed design at https://github.com/opendatahub-io/models-as-a-service/pull/1453

## Summary by CodeRabbit

- **Documentation**
  - Updated the approved architecture decision record for subscription-based flow control in model-serving requests.
  - Corrected the architecture decision record identifier to match its document title.
  - Standardized terminology around “inference priority” and the `inferencePriority` subscription setting.
  - Continues to document priority handling, fairness identifiers, authorization metadata, lifecycle and status behavior, operator controls, compatibility, security, deferred work, alternatives, risks, validation guidance, stakeholders, and references.


architecture-decision-records/model-serving/ODH-ADR-MS-0006-maas-flow-control.md
```diff
@@ -0,0 +1,192 @@
...
+* Change authorization, subscription selection, token rate limits, billing, or non-MaaS Gateway defaults.
+* Introduce `service_tier` overrides, per-user fairness, per-model priorities, or subscription-level scheduler policies.
+* Guarantee latency, minimum capacity, or global fairness across independent pools or schedulers.
+* Support flow control for external providers or routes without an `InferencePool`.
+
+## How
+
+### Subscription priority and tenant fairness
+
+Add optional `MaaSSubscription.spec.requestPriority`, a signed `int32` matching `InferenceObjective.spec.priority`.
```
mariusdanciu: A minor aspect. Instead of `requestPriority` name I'd suggest `inferencePriority` to be a little more explicit.

PR #171, opened 2026-09-22
Add ODH-ADR-Operator-0017: Shared Database Infrastructure Service

macgregor:
Introduces an ADR for a shared database infrastructure service: a declarative, Kubernetes-native way for
OpenShift AI components to request PostgreSQL access, modelled on `PersistentVolume` / `PersistentVolumeClaim`.
Administrators configure supply through `DatabaseProvider`; components express demand through `SchemaClaim` or
`DatabaseClaim`; the operator binds the two and writes a credentials `Secret` into the claim's namespace.

Originally drafted by @lburgazzoli, with a working
[proof of concept](https://github.com/lburgazzoli/opendatahub-module-operator/tree/db-service/modules/opendatahub-db-operator).
This revision hardens the connection surface against an inventory of the components that would actually
consume it, and settles the questions that draft left open. Tracked in
[RHOAIENG-94723](https://redhat.atlassian.net/browse/RHOAIENG-94723).

**On the number:** 0016 is claimed by #166 and 0015 by both `main` and #156, so this is 0017. Happy to
renumber to whatever falls out of merge order.

## The decisions worth your attention

These are the load-bearing ones. Each is argued in full in the document; they are listed here so review can go
straight at them.

1. **The credential `Secret` publishes discrete fields only — no composed URI or DSN.** Consumers build their
   own connection string. This is the most consequential decision in the ADR and the one most likely to draw
   an objection, so it is pre-empted deliberately in Alternatives rather than left to be relitigated. The cost
   is stated rather than hidden: four of the six named consumers must change code to adopt, and MLflow needs a
   CRD change where it would otherwise have adopted for free. The reasoning is that a published URI freezes a
   composition *algorithm* rather than a set of names, and there is no single URI syntax correct across the
   drivers in use.

2. **The service does not define, gate on, or enforce a PostgreSQL version range.** It reads the server
   version and reports it, and never blocks binding on it. This reverses an obvious expectation and partially
   redirects a commitment made in the RHAIRFE-1141 engineering response: the supported-version statement is
   real, but its subject is the product compatibility matrix, not one optional feature's API. The operator's
   own DDL needs nothing newer than PostgreSQL 9.3, so any floor asserted here would be invented.

3. **Both provider types ship in the initial version, and the default Internal provider is auto-created.**
   Neither is a stepping stone to the other. The Internal backend is deliberately basic — single instance, no
   HA, no automated backup/restore — and exists so the platform has a working default with zero administrator
   input. Anyone needing durability or compliance guarantees points an External provider at infrastructure
   that has them.

4. **CRD kind names are engine-neutral, with a `spec.engine` discriminator.** `DatabaseProvider`,
   `SchemaClaim`, `DatabaseClaim`, `DatabaseService` rather than `Postgres`-prefixed kinds. Kind names are
   immutable after release, so this is settled in the ADR rather than left to implementation. The cost is
   accepted plainly: a generic name over-promises while the service is PostgreSQL-only, and CEL validation
   rejects any other engine value so the over-promise is visible at admission.

5. **Claims declare the PostgreSQL extensions they require, in one provider-type-agnostic field.** On the
   Internal provider the operator installs them from a supported set; on an External provider it verifies and
   reports, never installing, because `CREATE EXTENSION` needs privileges the operator should not hold on
   infrastructure it does not own. The ADR fixes that mechanism but not the membership of the supported set,
   which is published per release alongside the PostgreSQL image it belongs to.

## One-way doors

Called out in their own section so review addresses them deliberately: the credential `Secret` key names,
the cluster-versus-namespace scope of each kind, the default provider name `rhai-db`, and the CRD kind names
above. Adding keys or extensions is additive and safe; renaming or removing is not.

## Deliberately deferred

Each is additive to re-open, so none of them constrains the initial version:

- **Bind-only claims** — binding to credentials that already exist, provisioning nothing. The case is served
  today by a hand-authored Connection `Secret`, which works because the ADR requires that consumers cannot
  distinguish a provisioned `Secret` from an authored one.
- **Idle reclaim of the Internal backend** — removed from the draft. It was the only path in the design that
  destroyed customer data on a timer, and no claims referencing a provider is not evidence the data is
  unwanted.
- **Credential rotation**, external secret-management integration, and non-PostgreSQL engines.

## Out of scope for this PR

Where the code ships — in-tree service controller, a module repository, or a standalone operator — is a
delivery decision tracked separately and does not belong in an API decision record. The choice of PostgreSQL
image for the Internal backend is likewise a supply-chain decision; the ADR commits only to a Red Hat
supported image.

## Relationship to existing ADRs

The output `Secret` conforms to the Connection API
([ODH-ADR-Operator-0009](https://github.com/opendatahub-io/architecture-decision-records/blob/main/architecture-decision-records/operator/ODH-ADR-Operator-0009-connection-api.md))
and carries `opendatahub.io/connection-type-protocol: "postgres"`. Registering that protocol with the
Connection API owners is a prerequisite of the design, not an optional enhancement. The ADR includes a section
explaining why the claim and provider resources warrant a new API group rather than reusing that one, and an
Alternatives entry rejecting full reuse.

[RHOAIENG-94723]: https://redhat.atlassian.net/browse/RHOAIENG-94723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Documentation**
  - Added an architecture decision record defining a shared PostgreSQL database infrastructure service for OpenShift AI.
  - Documented resource types for database providers, services, schema claims, and database claims.
  - Defined credential delivery through Kubernetes Secrets, including connection details and TLS support.
  - Documented support for approved PostgreSQL extensions and external or platform-managed database providers.
  - Recorded decisions around credential rotation, resource reclamation, versioning, and data migration.


lburgazzoli:
@danielezonca @mariusdanciu @lphiri @abiazett @LukaszCmielowski mind having a look ?

architecture-decision-records/operator/ODH-ADR-Operator-0017-shared-database-infrastructure-service.md
```diff
@@ -0,0 +1,605 @@
...
+> **A consuming component accepts a `postgres` Connection `Secret`, and must not be able to tell whether it was produced by a claim or written by an administrator.**
+
+This is what allows an administrator to point a component at a database this service never provisioned — a pre-existing corporate PostgreSQL, a cloud managed instance, an instance owned by another team — without the component needing a second configuration path, and without the platform needing admin credentials on that database. It is also what lets a component adopt the standard contract *before*, or entirely without, the provisioning side of this service being enabled.
+
+**The output `Secret` therefore conforms to the Connection API,** carrying:
+
+```yaml
+metadata:
+  annotations:
+    opendatahub.io/connection-type-protocol: "postgres"
```
mariusdanciu: This won't be relevant for new connections API (DCH). 

But I wonder why the Connection API is relevant here. Connections use by users to configure storage. I was under the impression for this ADR is for RHOAI own services to be able to request a separate DB or schema but I could be wrong.

architecture-decision-records/operator/ODH-ADR-Operator-0017-shared-database-infrastructure-service.md
```diff
@@ -0,0 +1,605 @@
...
+
+Drift recovery is **not** credential rotation, and this ADR does not introduce rotation. A password changes only when database-side state has to be re-provisioned; there is no scheduled or on-demand rotation workflow. Rotation remains a Non-Goal, and delivery work that assumes otherwise should be cut back to match.
+
+### One-Way Doors
+
+Three decisions in this design cannot be reversed after release without breaking existing installations. They are called out explicitly so that review addresses them deliberately.
+
+1. **Credential `Secret` key names.** Every adopting component encodes them. Changing a key after release breaks every adopter simultaneously. Adding keys is safe; renaming or removing is not. Note that the decision *not* to publish a composed URI keeps this door as narrow as it can be: a published URI would additionally freeze a composition **algorithm**, and an algorithm is a considerably worse thing to be unable to change than a name. Adding a URI key later remains possible; removing one would not have been.
```
LukaszCmielowski: Good point. Is there any single source of truth today ?. @mariusdanciu can DCH be used as such ( connection type is well defined including the secret schema) ?.

mariusdanciu: DCH can point to any secret. However our Postgres connection type defines `URI` and `CA_CERT` fields. But see my note above https://github.com/opendatahub-io/architecture-decision-records/pull/171#discussion_r4093741634. I am not sure that users would define connection pointing to secrets coming from these claims. These DBs are mainly used for internal RHOAI service that require storage not so much for user data storage - at least this is my understanding.

lburgazzoli: correct, this is not a generic database service, it is for RHAI services

